PR #3, opened 2021-05-06
APM-293899 set up Travis pipeline

wojtasskorcz:


.travis.yml
```diff
@@ -0,0 +1,46 @@
...
+      # using 'script' instead of 'before_deploy' because it's unskippable in Python environment
+      # https://docs.travis-ci.com/user/languages/python/#default-build-script
+      script:
+        - chmod +x build-release-package.sh
+        - chmod +x version.sh
+        - ./build-release-package.sh # needs Python environment internally
+      deploy:
+        provider: releases
+        api_key: $GITHUB_RELEASE_API_KEY
+        file: ./dynatrace-aws-log-forwarder.zip
```
wojtasskorcz, line 43: I opted for a single release package (unlike Azure) because:

- It doesn't require CloudFormation template modifications in Travis, see: https://github.com/dynatrace-oss/dynatrace-azure-log-forwarder/blob/master/.travis.yml#L41
- It works even after repo rename or any other changes to the URL of the remote files
- `aws lambda update-function-code` doesn't support remote GH files, only local ones or S3

.travis.yml
```diff
@@ -0,0 +1,36 @@
+language: shell
+notifications:
+  email:
+    on_success: never
+stages:
+- tests
+- deployment
+env:
+  global:
+    - RELEASE_REPO_TO_FIND="https://github.com/dynatrace-oss/dynatrace-aws-log-forwarder/releases/download/latest/"
```
wojtasskorcz, line 10: remove

pipeline/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM public.ecr.aws/sam/build-python3.8
```
wojtasskorcz, line 1: use `amazon/aws-sam-cli-build-image-python3.8`

.travis.yml
```diff
@@ -0,0 +1,36 @@
...
+        - docker run --rm lambda_image bash -c 'python3 -m pylint --rcfile=pipeline/pylint.cfg $(find src/ -name "*.py")'
+        - docker run --rm lambda_image bash -c 'cd src && python3 -m pytest -v ../tests/unit'
+    - stage: deployment
+      name: Github Release Deployment
+      if: tag =~ /^release.*$/
+      language: shell
+      services: docker
+      script: >
+        docker run --rm -v `pwd`:/var/task amazon/aws-sam-cli-build-image-python3.8 bash -c
+        'chmod +x build-release-package.sh && chmod +x version.sh && chmod +x dynatrace-aws-logs.sh && ./build-release-package.sh'
```
wojtasskorcz, line 29: try to remove chmod (keep +x in git)

mmajcher (approved):

.travis.yml
```diff
@@ -0,0 +1,36 @@
...
+      name: Tests & Linting
+      language: shell
+      services: docker
+      script:
+        - docker build -f pipeline/Dockerfile -t lambda_image .
+        - docker run --rm lambda_image bash -c 'python3 -m pylint --rcfile=pipeline/pylint.cfg $(find src/ -name "*.py")'
+        - docker run --rm lambda_image bash -c 'cd src && python3 -m pytest -v ../tests/unit'
+    - stage: deployment
+      name: Github Release Deployment
+      if: tag =~ /^release.*$/
```
dt-be, line 20: summary from what you explained: new tag "release.*" triggers entire travis build + as this condition is true, also release package is created and lands in here:
https://github.com/dynatrace-oss/dynatrace-aws-log-forwarder/releases

Afterwards one-liner should be adapted to use latest release

.travis.yml
wojtasskorcz: it worked! :)

version.sh
```diff
@@ -0,0 +1,36 @@
...
+#     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#     See the License for the specific language governing permissions and
+#     limitations under the License.
+
+echo  "Travis tag     '$TRAVIS_TAG'"
+echo  "Travis commit  '$TRAVIS_COMMIT'"
+
+version=""
+
+if [ -n "$TRAVIS_TAG" ]
```
dt-be, line 21: summary from what you explained: 
- these are Travis env vars
- as this version.sh is used only when during "deployment" stage/building the release zip, which is triggered by adding "release.*" tag, the TRAVIS_TAG should be there, and this is just in case

.travis.yml
```diff
@@ -0,0 +1,36 @@
+language: shell
+notifications:
+  email:
+    on_success: never
```
dt-be, line 4: Personally I'd opt for having an email about new releases (not about every build)

PR #7, opened 2021-05-11
Feature/apm 290579 implement self monitoring

mmajcher:
Live example 

https://console.aws.amazon.com/cloudwatch/home?region=us-east-1#dashboards:name=DynatraceLogForwarder-SelfMonitoring-us-east-1-dynatrace-aws-logs;start=PT1H

account 9080


template-aws-logs.yaml
```diff
@@ -141,6 +149,323 @@ Resources:
...
+                      "height": 6,
+                      "width": 12,
+                      "y": 6,
+                      "x": 12,
+                      "type": "metric",
+                      "properties": {
+                          "metrics": [
+                              [ "DT/LogsStreaming", "Kinesis record age", "function_name", "${__LambdaName__}", { "stat": "Minimum" } ],
+                              [ "...", { "stat": "Average" } ],
+                              [ "..." ]
```
wojtasskorcz, line 173: Shouldn't it be `{ "stat": "Average" }` here?

template-aws-logs.yaml
```diff
@@ -141,6 +149,323 @@ Resources:
...
+                      "height": 6,
+                      "width": 12,
+                      "y": 0,
+                      "x": 0,
+                      "type": "metric",
+                      "properties": {
+                          "view": "timeSeries",
+                          "stacked": false,
+                          "metrics": [
+                              [ "AWS/Lambda", "Invocations", "FunctionName", "BeluAWSLogsPerformanceTest_inputLogsProducingLambda" ]
```
wojtasskorcz, line 442: this and below is fishy

template-aws-logs.yaml
mmajcher: If this is not specified then it's taken from higher level "stat", there is stat=Maximum 5 lines below

template-aws-logs.yaml
```diff
@@ -141,6 +149,323 @@ Resources:
...
+                  {
+                      "height": 6,
+                      "width": 12,
+                      "y": 42,
+                      "x": 12,
+                      "type": "metric",
+                      "properties": {
+                          "metrics": [
+                              [ "DT/LogsStreaming", "Requests duration", "function_name", "${__LambdaName__}", { "stat": "Minimum" } ],
+                              [ "..." ],
```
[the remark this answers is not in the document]
mmajcher: Same as previous example, it takes stat=Average from higher level.
This is auto-generated from Cloudwatch and I have no idea how the "main stat" (level up in this object) is decided. Anyhow, it does work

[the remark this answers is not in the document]
wojtasskorcz: Interesting, thanks :)

wojtasskorcz (approved):

wojtasskorcz (approved):

src/logs/logs_sender.py
```diff
@@ -92,38 +109,40 @@ def prepare_serialized_batches(logs: List[Dict]) -> List[str]:
...
-def ensure_content_length(log_entry):
+def ensure_content_length(log_entry, context):
     log_content = log_entry.get("content", "")
     log_content_len = len(log_content)
     if log_content_len > DYNATRACE_LOG_INGEST_CONTENT_MAX_LENGTH:
         # placeholder for APM-290579 Implement self-monitoring
         # dynatrace-azure-logs-ingest/logs_ingest/main.py:117
         # self_monitoring.too_long_content_size.append(log_content_len)
 
+        context.sfm.log_content_trimmed()
```
dt-be, line 138: Please clean the placeholder above.

+ i'd rather report what was the original length that bool if required trimming. It seems more informative to me, e.g., if it's a problem for the customer, he might be able to manipulate the limit in Dynatrace cluster. 
(Admittedly, there is little chance of this happening unless we document it, but it still seems a step in the right direction)

wojtasskorcz: Would you like to sum the length of trimmed logs when sending a metric datapoint to CW? Because then I see no added value for the customer. If not, how would we report it to CloudWatch?

dt-be (approved):

src/logs/logs_sender.py
wojtasskorcz: Well, then we should report it *in addition* to the number of trimmed logs because it's also important.

Anyway, I think we can add it in the future if customers ask for it (similarly to how we added region reporting in metric streaming when the need arrived) but right now I'm against prolonging the work on this PR, especially since the comment comes two days after Marek initially presented it to us. I think it's asking a bit too much to add non-essential stuff after such a long period

PR #9, opened 2021-05-18
Fix description of --role-arn script parameter

wojtasskorcz:
See `--role-arn` description in https://docs.aws.amazon.com/cli/latest/reference/logs/put-subscription-filter.html


dt-be:
You might want to rename that one, too:

dynatrace-aws-log-forwarder-template.yaml >RoleForSubscriptionFilters

dt-be (approved):

PR #10, opened 2021-05-21
APM-283479 Performance test dashboard - Lambda errors & Total decompressed

dt-be:
Added widget for Lambda errors & Total decompressed MB

After a lot of second and third thoughts I've kept my reordering - for me it's better for debugging issues, previous 'organized in sections' structure imho doesn't and is more about esthetics than practicality. Other opinions welcome, I can reorder it back and squeeze the new chart somewhere, if you prefer.

You can compare the dashboards live in our 90... account, or the screenshots below:

Before:
image
After:
image

### DIFF
--- a/dynatrace-aws-log-forwarder-template.yaml
+++ b/dynatrace-aws-log-forwarder-template.yaml
@@ -467,7 +467,7 @@ Resources:
               "type": "metric",
               "properties": {
                 "metrics": [
-                [ "...", "${__LambdaName__}" ]
+                [ "AWS/Lambda", "Errors", "FunctionName", "${__LambdaName__}" ]
                 ],
                 "view": "timeSeries",
                 "stacked": false,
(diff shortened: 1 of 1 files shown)


dynatrace-aws-log-forwarder-template.yaml
```diff
@@ -173,311 +173,347 @@ Resources:
...
+            },
+            {
+              "height": 6,
+              "width": 12,
+              "y": 12,
+              "x": 12,
+              "type": "metric",
+              "properties": {
+                "metrics": [
+                [ "...", "${__LambdaName__}" ]
```
wojtasskorcz, line 470: How is namespace and metric name determined in this case?

dt-be: Thanks, fixing, it seems I used to have there 2 same metrics for different functions, and ate up the part that actually defined those:

"metrics": [
                    [ "AWS/Lambda", "Errors", "FunctionName", "BeluAWSLogsPerformanceTest_inputLogsProducingLambda", { "visible": false } ],
                    [ "...", "BeluPerfLogsCLient-Lambda-1VXEVIDV8B4TC" ]
                ],

wojtasskorcz (approved):

PR #12, opened 2021-05-25
APM-299788 APM-293899 one-liner & Travis - no cd dir in one-liner

dt-be:
The version before ended me up with multiple levels of directories + it kept changing the directory I'm in every time I called the one-liner. 

I called it again and again to try to download new (test) release, but (and this is customer use case to me) also whenever I had some mistake in the setting of env_vars. Ofc, it would be best to run just the deploy, and spare yourself/github the wget, but this requires you to dive into the content of the one-liner, rather than a no-brainer of running the same command again.

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ REQUIRE_VALID_CERTIFICATE=false
 Download the script and deploy the infrastructure:
 
 ```
-wget https://github.com/dynatrace-oss/dynatrace-aws-log-forwarder/releases/latest/download/dynatrace-aws-log-forwarder.zip && unzip -q dynatrace-aws-log-forwarder.zip && cd dynatrace-aws-log-forwarder && ./dynatrace-aws-logs.sh deploy --target-url $TARGET_URL --target-api-token $TARGET_API_TOKEN --require-valid-certificate $REQUIRE_VALID_CERTIFICATE
+wget -O dynatrace-aws-log-forwarder.zip https://github.com/dynatrace-oss/dynatrace-aws-log-forwarder/releases/latest/download/dynatrace-aws-log-forwarder.zip && unzip -qo dynatrace-aws-log-forwarder.zip && ./dynatrace-aws-logs.sh deploy --target-url $TARGET_URL --target-api-token $TARGET_API_TOKEN --require-valid-certificate $REQUIRE_VALID_CERTIFICATE
 ```
 
 #### Usage and options reference - deploy
--- a/build-release-package.sh
+++ b/build-release-package.sh
@@ -48,6 +48,8 @@ rm -rf $LAMBDA_BUILD_DIR
 mkdir $PACKAGE_BUILD_DIR
 
 cp README.md dynatrace-aws-logs.sh $LAMBDA_ZIP_NAME dynatrace-aws-log-forwarder-template.yaml $PACKAGE_BUILD_DIR
-zip $PACKAGE_ZIP_NAME $PACKAGE_BUILD_DIR/*
+cd $PACKAGE_BUILD_DIR
+zip ../$PACKAGE_ZIP_NAME *
+cd ..
 
 rm -rf $PACKAGE_BUILD_DIR


build-release-package.sh
```diff
@@ -48,6 +48,8 @@ rm -rf $LAMBDA_BUILD_DIR
 mkdir $PACKAGE_BUILD_DIR
 
 cp README.md dynatrace-aws-logs.sh $LAMBDA_ZIP_NAME dynatrace-aws-log-forwarder-template.yaml $PACKAGE_BUILD_DIR
-zip $PACKAGE_ZIP_NAME $PACKAGE_BUILD_DIR/*
+cd $PACKAGE_BUILD_DIR
+zip ../$PACKAGE_ZIP_NAME *
+cd ..
```
dt-be, line 53: before the structure was:
dynatrace-aws-log-forwarder.zip/dynatrace-aws-log-forwarder/<files>
now it is:
dynatrace-aws-log-forwarder.zip/<files>

so that the unzip doesn't produce unzip dynatrace-aws-log-forwarder/<files> from the zip, but only <files> directly to the current dir

I'm more fine with 4 files spamming user's dir than with secretly spamming the user with multiple levels of dirs lower and lower.

README.md
```diff
@@ -31,7 +31,7 @@ REQUIRE_VALID_CERTIFICATE=false
 Download the script and deploy the infrastructure:
 
 ```
-wget https://github.com/dynatrace-oss/dynatrace-aws-log-forwarder/releases/latest/download/dynatrace-aws-log-forwarder.zip && unzip -q dynatrace-aws-log-forwarder.zip && cd dynatrace-aws-log-forwarder && ./dynatrace-aws-logs.sh deploy --target-url $TARGET_URL --target-api-token $TARGET_API_TOKEN --require-valid-certificate $REQUIRE_VALID_CERTIFICATE
+wget -O dynatrace-aws-log-forwarder.zip https://github.com/dynatrace-oss/dynatrace-aws-log-forwarder/releases/latest/download/dynatrace-aws-log-forwarder.zip && unzip -qo dynatrace-aws-log-forwarder.zip && ./dynatrace-aws-logs.sh deploy --target-url $TARGET_URL --target-api-token $TARGET_API_TOKEN --require-valid-certificate $REQUIRE_VALID_CERTIFICATE
```
dt-be, line 34: wget -O is what one-agent one-liner also does, to make sure the zip get overwritten on each call of the wget, rather than create a dynatrace-aws-log-forwarder.zip.1, dynatrace-aws-log-forwarder.zip.2 that will not be used by the unzip

unzip -o allows overwritting the target files without asking about each one of them. I found it annoying to have to click y(es) four times, but we can talk about it. Excerpt from one-liner output without -o below:
image

build-release-package.sh
wojtasskorcz: This bears some signs of a tarbomb but I guess four files is not that bad really, especially since they should have names different than other files in the user's directory and so shouldn't overwrite anything. Fine by me

README.md
wojtasskorcz: Good catch! I haven't checked the overwrite-on-unzip scenario and didn't notice that before

wojtasskorcz (approved):

README.md
dt-be: Tested with one-liner from this PR with a slight alteration for custom release, works as planned:
```
#RELEASE="latest"
RELEASE="release-test-new-zip-and-oneliner"
GITHUB_RELEASE=https://github.com/dynatrace-oss/dynatrace-aws-log-forwarder/releases/download/$RELEASE/dynatrace-aws-log-forwarder.zip
wget -O dynatrace-aws-log-forwarder.zip $GITHUB_RELEASE && unzip -qo dynatrace-aws-log-forwarder.zip && ./dynatrace-aws-logs.sh deploy --target-url $TARGET_URL --target-api-token $TARGET_API_TOKEN --require-valid-certificate $REQUIRE_VALID_CERTIFICATE
```
image

PR #15, opened 2021-06-10
Apm 299389 metadata extraction and entity mapping for cloudtrail rds

dt-be:
~~open for initial comments on the config and its readability when there are more services in the CloudTrail file.~~

REVIEW READY

https://github.com/dynatrace-oss/dynatrace-azure-log-forwarder/blob/master/logs_ingest/config/activity_logs.json
https://github.com/dynatrace-oss/dynatrace-gcp-function/blob/d434130a50fac7889f61ee380cadeb411eaea58b/src/config_logs/cloud_audit_logs.json


src/logs/metadata_engine/metadata_engine.py
```diff
@@ -139,15 +140,22 @@ def _apply_rule(rule, record, parsed_record):
     if rule.aws_loggroup_pattern and "log_group" in record:
         extracted_values = parse_aws_loggroup_with_grok_pattern(record["log_group"], rule.aws_loggroup_pattern)
         record.update(extracted_values)
+    if rule.log_content_parse_type in "json":
```
mmajcher, line 143: "in" or "is"?

src/config/cloudtrail.json
```diff
@@ -0,0 +1,57 @@
+{
+  "name": "CLOUD_TRAIL",
+  "displayName": "Cloud Trail Management Events / Logs",
+  "rules": [
+    {
+      "aws": {
+        "logGroup": "aws-cloudtrail-logs-%{DATA:account_id}-%{GREEDYDATA:suffix}",
```
mmajcher, line 7: do we need account_id in log group? I think we get this from Kinesis in batch

src/config/cloudtrail.json
```diff
@@ -0,0 +1,57 @@
...
+
+          "pattern": "[ { condition_field: errorCode, result: [`Failure`, errorCode] | [? @!=null ] | join('.', @) }, { condition_field: `defaultValue`, result: 'Succeeded'} ] | [?condition_field != null && condition_field !=``] | [0] | result"
+        },
+
+
+        {
+          "key": "dt.source_entity_type",
+          "description": "If a field given in condition_field exists and matches the condition value, use the value from result.",
+
+          "pattern": "[ { condition: `rds.amazonaws.com`, condition_field: log_content_parsed.eventSource, result: 'RELATIONAL_DATABASE_SERVICE'} ] | [?condition_field == condition] | [0] | result"
```
mmajcher, line 42: Wow, that's nice 👍

dt-be: Quite proud of it, too. 
Although, the final condition might need to be on a different field. I think there can be more events in that eventSource=rds that are not-instance bound - I still need to get back to that.

src/logs/metadata_engine/metadata_engine.py
dt-be: image

dt-be: But in was a bad idea, it crashes with left hand side =None

src/config/cloudtrail.json
```diff
@@ -29,27 +29,130 @@
...
+          "key": "dt.source_entity",
+          "pattern": "if( log_content_parsed.eventName == 'RebootDBInstance', &dt_meid_rds_v2( format('arn:aws:rds:{}:{}:db:{}', region, account_id, log_content_parsed.requestParameters.dBInstanceIdentifier)), &EmptyValue, @ )"
+        },
+        {
+          "key": "dt.source_entity",
+          "pattern": "if( log_content_parsed.eventName == 'DescribeEvents', &EmptyValue, &EmptyValue, @ )"
+        },
+        {
+          "key": "dt.source_entity",
+          "pattern": "if( log_content_parsed.eventName == 'RebootDBInstance', &dt_meid_rds_v2( format('arn:aws:rds:{}:{}:db:{}', region, account_id, log_content_parsed.requestParameters.dBInstanceIdentifier)), &EmptyValue, @ )"
```
mmajcher, line 145: 1)I think that we should probably introduce "switch" instruction to jmespath.
2) One thing to check is where arguments evaluation happen. To avoid evaluating all arguments to if/switch this evaluation should preferably be done only after condition is true. But this might be asking for too much, right? 
Otherwise each if's arguments gets evaluated? Lot's of unnecessary calculations.

src/logs/metadata_engine/metadata_engine.py
```diff
@@ -227,7 +227,7 @@ def _create_config_rule(entity_name: str, rule_json: Dict) -> Optional[ConfigRul
         aws_loggroup_pattern = rule_json["aws"]["logGroup"]
     except KeyError:
         aws_loggroup_pattern = None
-    log_content_parse_type = rule_json.get("aws", {}).get("logContentParseAs", None)
+    log_content_parse_type = rule_json.get("aws", {}).get("logContentParseAs", "")
```
mmajcher, line 230: Why this change, given that you are checking this field with "==" ?

src/config/cloudtrail.json
```diff
@@ -29,27 +29,130 @@
...
-          "key": "dt.source_entity_type",
-          "description": "If a field given in condition_field exists and matches the condition value, use the value from result.",
+          "key": "dt.source_entity",
+          "pattern": "if( log_content_parsed.eventName == 'DescribeOptionGroups', &EmptyValue, &EmptyValue, @ )"
+        },
 
-          "pattern": "[ { condition: `some other source of cloudTrail log event`, condition_field: log_content_parsed.eventSource, result: 'some other entity, e.g. CUSTOM_DEVICE'} ] | [?condition_field == condition] | [0] | result"
+        {
+          "key": "dt.source_entity",
+          "pattern": "if( log_content_parsed.eventName == 'DescribeOrderableDBInstanceOptions', &EmptyValue, &EmptyValue, @ )"
```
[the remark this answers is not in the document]
dt-be: Temp to know what I've done and what not, I pushed it to have a real release for demo.

src/config/cloudtrail.json
dt-be: 1. I'm against the switch as mentioned on demo, at least the way I understand what you suggest (=big single rule with one switch for them all, line that cannot be broken or formatted). If you don't agree let's have a call.
2. As mentioned on demo, & makes it late resolved. If you have doubts, I can demo in on a call.

mmajcher: 1. Yes, big single rule, is it not possible to have this with multiple lines??
2. Great

dt-be: 1. no, no line breaks within a string in json. Sad us during the hackaton.

mmajcher: Oh right. 😞

src/logs/metadata_engine/metadata_engine.py
dt-be: image
Hmm, it made it harder for me to make local debugging changes of the sort of:
```
record["log_content_parsed"]["requestParameters"] = {
        "dBInstanceIdentifier": "belu-metadata-database-1"
    }
```
as log_content_parsed wasn't there. 

Not sure if in python check against None or empty string makes a lot of difference. I'd say not and I'd leave it, but if you want it back, I won't insist.

dt-be: Must have appeared in the process of switching to ==

src/config/cloudtrail.json
dt-be: Switched to ifs

src/config/cloudtrail.json
```diff
@@ -0,0 +1,57 @@
...
+        },
+        {
+          "key": "audit.identity",
+          "pattern": "log_content_parsed.userIdentity.arn"
+        },
+        {
+          "key": "audit.status",
+          "description":  "If a field errorCode exists, then return Failure.{errorCode}, else return Succeeded. [? @!=null] filters out null values, as otherwise the join method would crash when there was no errorCode",
+
+          "pattern": "[ { condition_field: errorCode, result: [`Failure`, errorCode] | [? @!=null ] | join('.', @) }, { condition_field: `defaultValue`, result: 'Succeeded'} ] | [?condition_field != null && condition_field !=``] | [0] | result"
```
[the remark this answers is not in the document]
dt-be: Switched to ifs, they were in this project, too ;)

src/logs/metadata_engine/metadata_engine.py
dt-be: tested & reverted

src/logs/metadata_engine/metadata_engine.py
```diff
@@ -223,10 +223,7 @@ def _create_config_rule(entity_name: str, rule_json: Dict) -> Optional[ConfigRul
         return None
     attributes = _create_attributes(rule_json.get("attributes", []))
 
-    try:
-        aws_loggroup_pattern = rule_json["aws"]["logGroup"]
-    except KeyError:
-        aws_loggroup_pattern = None
+    aws_loggroup_pattern = rule_json.get("aws", {}).get("logGroup", None)
```
dt-be, line 226: https://docs.python.org/2/faq/design.html#how-fast-are-exceptions

src/config/cloudtrail.json
```diff
@@ -0,0 +1,753 @@
...
+          "pattern": "log_content_parsed.eventName"
+        },
+        {
+          "key": "audit.identity",
+          "pattern": "log_content_parsed.userIdentity.arn"
+        },
+        {
+          "key": "audit.result",
+          "description": "If a field errorCode exists, then return Failure.{errorCode}, else return Succeeded",
+          "pattern": "if( log_content_parsed.errorCode==null, &'Succeeded', &[`Failed`, log_content_parsed.errorCode] | [? @!=null ] | join('.', @) , @)"
```
wojtasskorcz, line 32: Is `[? @!=null ]` neede here? We already checked that `log_content_parsed.errorCode` is not null

src/config/cloudtrail.json
```diff
@@ -0,0 +1,753 @@
...
+          "key": "severity",
+          "pattern": "if( log_content_parsed.errorCode==null, &'INFO', &'ERROR', @)"
+        },
+        {
+          "key": "dt.source_entity",
+          "pattern": "if( log_content_parsed.eventName == 'AddRoleToDBCluster', &dt_meid_supporting_service_v2( 'Aurora', format('arn:aws:rds:{}:{}:cluster:{}', [region, account_id, log_content_parsed.requestParameters.dBClusterIdentifier])), &EmptyValue, @ )"
+        },
+        {
+          "key": "aws.arn",
+          "pattern": "if( log_content_parsed.eventName == 'AddRoleToDBCluster', &format('arn:aws:rds:{}:{}:cluster:{}', [region, account_id, log_content_parsed.requestParameters.dBClusterIdentifier]), &EmptyValue, @ )"
```
wojtasskorcz, line 44: Do I remember correctly that you wanted to try to avoid repeating the ARN calculation in `dt.source_entity` and `aws.arn`? Is it still your plan?

src/config/cloudtrail.json
```diff
@@ -0,0 +1,753 @@
...
+        {
+          "key": "aws.resource.id",
+          "pattern": "if( log_content_parsed.eventName == 'AddRoleToDBInstance', &log_content_parsed.requestParameters.dBInstanceIdentifier, &EmptyValue, @ )"
+        },
+        {
+          "key": "dt.source_entity",
+          "pattern": "if( log_content_parsed.eventName == 'AddSourceIdentifierToSubscription', &dt_meid_rds_v2( format('arn:aws:rds:{}:{}:db:{}', [region, account_id, log_content_parsed.requestParameters.dBInstanceIdentifier])), &EmptyValue, @ )"
+        },
+        {
+          "key": "dt.source_entity",
```
wojtasskorcz, line 67: `dt.source_entity` set twice for the same `log_content_parsed.eventName` -- is it intentional?

src/logs/metadata_engine/jmespath.py
```diff
@@ -44,14 +44,23 @@ def _func_if(self, condition, if_true_expression, if_false_expression, node_scop
         else:
             return if_false_expression.visit(if_false_expression.expression, node_scope)
 
+
+    @functions.signature({'types': ['string', 'null']}, {'types': ['string', 'null']})
+    def _func_starts_with(self, search, suffix):
```
wojtasskorcz, line 49: `prefix` instead of `suffix`

src/logs/metadata_engine/jmespath.py
```diff
@@ -66,4 +75,62 @@ def _func_dt_meid_md5(self, entity_type, hashing_input):
...
+    def _func_dt_meid_supporting_service_v1(self, supporting_service_name, region, main_dimension, name):
+        custom_device_input_string = "-".join([supporting_service_name.lower(), region, main_dimension, name])
+
+        # out.writeLong(customDeviceGroupLongId);
+        # out.write(customDeviceIdBytes);
+        # out.writeInt(customDeviceIdBytes.length);
+        #
+        # return MURMUR_128.hashBytes(byteOut.toByteArray()).asLong();
+
+        raise NotImplementedError
```
wojtasskorcz, line 111: Maybe we should add a comment here that it cannot be implemented because it needs Credentials ID that is generated randomly by the AG? Also, what do the commented lines above mean?

tests/unit/logs/test_transformation_full.py
```diff
@@ -98,7 +97,175 @@
...
+                                                   "eventType": "AwsApiCall",
+                                                   "managementEvent": True,
+                                                   "eventCategory": "Management",
+                                                   "recipientAccountId": "444000444"
+                                               })
+                                               }
+                                              ]},
+        "expect_first_log_contains": {
+            'aws.service': 'cloudtrail',
+            # "aws.resource.id": "TODO",
```
wojtasskorcz, line 180: TODOs left here and below

src/config/cloudtrail.json
dt-be: log_content_parsed.errorCode = null
[`Failed`, log_content_parsed.errorCode] | [? @!=null ] | join('.', @) > "Failed"
[`Failed`, log_content_parsed.errorCode] | join('.', @) > crash, TypeError: join() expected argument 2 to be type array-string but received type array instead.

For jmespath string and null (and so the arrays thereof) are different types but they are not both allowed in most jmespath methods definitions (only string). I could overwrite join the way I ended up overwriting built-in starts_with if you want - I just did both things differently at different stages.

dt-be: Btw, just to have it said out loud, I decided to translate
{  "log_content_parsed": 
    {"errorCode": null} }
as Failed (same for empty string, even though there is no errorCode. There is another field of errorMessage, that could still be not empty. As a user, I'd rather have an extra false positive Failed message, that I could clarify by looking at the log, rather than have a Failed hide somewhere in the ocean of Succeeded. 

I could check on errorMessage field, too, but that seems an edge case that didn't deserve expanding the pattern.

Btw, one more thing
log_content_parsed.errorCode = ""
[`Failed`, log_content_parsed.errorCode] | [? @!=null ] | join('.', @) > "Failed."

but we could potentially filter that out if you want:
[`Failed`, log_content_parsed.errorCode] | [? @!=null && @!=``] | join('.', @) > "Failed"

Btw, Azure and GCP also has Succeeded/Failed.somethingMore. Not sure if I checked the cases above against them.

src/config/cloudtrail.json
dt-be: I did, but I had trouble taming the chaos and had to write it all out plainly first before seeing the patterns and deciding what to do about it. I might need some follow up discussion about the options here.

src/logs/metadata_engine/jmespath.py
dt-be: I saw it, I can fix it, but this is and exact clone from original jmespath implementation, apart from null's appearing.

src/logs/metadata_engine/jmespath.py
dt-be: This is for cleanup imho. Commented code is a copy paste from AG code, when I was not 100% sure if I would be implementing it later or not.

tests/unit/logs/test_transformation_full.py
dt-be: Who said it's review ready? ;) that one still waits for commit&push

src/config/cloudtrail.json
wojtasskorcz:
```
log_content_parsed.errorCode = null
[Failed, log_content_parsed.errorCode] | [? @!=null ] | join('.', @) > "Failed"
[Failed, log_content_parsed.errorCode] | join('.', @) > crash, TypeError: join() expected argument 2 to be type array-string but received type array instead.
```
Will this scenario ever happen though? The expression should evaluate the on-true argument here (`&'Succeeded'`) and not evaluate the on-false argument. Or is my assumption about short-circuit evaluation wrong in JMESPath?

src/logs/metadata_engine/jmespath.py
wojtasskorcz: In their docs they have it named correctly: https://jmespath.org/specification.html#starts-with
I think we should name it correctly even if their implementation has got it wrong

dt-be:
Please review, it's as ready as it will be.

src/logs/metadata_engine/jmespath.py
```diff
@@ -44,26 +44,96 @@ def _func_if(self, condition, if_true_expression, if_false_expression, node_scop
...
     def _func_dt_meid_murmurhash(self, entity_type, hashing_input):
         return me_id.meid_murmurhash(entity_type, hashing_input)
 
+    #same ids across all version of aws credentials - md5
+    @functions.signature({'types': ['string', 'null']})
+    def _func_dt_meid_ebs_volume(self, volumeId): return me_id.meid_md5("EBS_VOLUME", volumeId)
+    @functions.signature({'types': ['string', 'null']})
+    def _func_dt_meid_ec2_instance(self, instanceId): return me_id.meid_md5("EC2_INSTANCE", instanceId)
+    @functions.signature({'types': ['string', 'null']},{'types': ['string', 'null']},{'types': ['string', 'null']})
+    def _func_dt_meid_lambda_function(self, functionName, region, accountId): return me_id.meid_md5("AWS_LAMBDA_FUNCTION", functionName, region, "_", accountId)
```
mmajcher, line 94: I feel you should prepare hashing input as a one string here so
region, "_", accountId -> region+"_"+accountId 
And not do any concatenations in hashing methods. 
But this is minor, feel free to ignore

src/logs/metadata_engine/me_id.py
```diff
@@ -19,19 +19,45 @@
...
     return identifier
 
+def meid_murmurhash_awsseed(entity_type: str, *hashing_input: str) -> str:
+    if None in hashing_input:
+        # any None value in input is treated as missing required data, no ID can be calculated
+        return None
+
+    concatented_input = "".join(hashing_input)
+
+    long_id = _murmurhash2_64A(concatented_input, seed = -512093083)
```
mmajcher, line 52: Can we have hex here so it is easier to find corresponding thing on cluster side? I mean for the seed

src/logs/metadata_engine/metadata_engine.py
```diff
@@ -115,7 +118,7 @@ def _load_configs(self):
                         self.default_rule = _create_config_rules(config_json)[0]
                     else:
                         self.rules.extend(_create_config_rules(config_json))
-            except Exception:
+            except Exception as ex:
                 logging.exception(f"Failed to load configuration file: '{config_file_path}'")
```
mmajcher, line 122: ex not added to log here

src/logs/metadata_engine/metadata_engine.py
```diff
@@ -127,8 +130,8 @@ def apply(self, record: Dict, parsed_record: Dict):
             # No matching rule has been found, applying the default rule
             if self.default_rule:
                 _apply_rule(self.default_rule, record, parsed_record)
-        except Exception:
-            logging.exception(f"Encountered exception when running Rule Engine")
+        except Exception as ex:
+            logging.exception(f"Encountered exception when running Rule Engine. ")
```
mmajcher, line 134: Same, ex can be logged? If not logged than "as ex" not needed. I don't know the reason why metadata engine does not log exceptions but there might be some. Theoretically there could be lots of exceptions, one exception for one log could kill your cloudwatch bill. Maybe some logging with throttling would be nice. I don't think you should do it here though

src/logs/metadata_engine/metadata_engine.py
```diff
@@ -139,14 +142,27 @@ def _apply_rule(rule, record, parsed_record):
...
     for attribute in rule.attributes:
         try:
             value = jmespath.search(attribute.pattern, record, JMESPATH_OPTIONS)
             if value:
                 parsed_record[attribute.key] = value
-        except Exception:
-            logging.exception(f"Encountered exception when evaluating attribute {attribute} of rule for {rule.entity_type_name}")
+
+                # attributes with priority are available for the calculation of further attributes
+                if attribute.priority is not None:
```
mmajcher, line 160: So "priority" only means that it goes (as usual) to output but also additionally to input? 
Priority kinda suggests reordering? But we don't have it. So it's more like "allowReferencing" or something? Just thinking, not strong about this in any way

src/logs/metadata_engine/metadata_engine.py
```diff
@@ -139,14 +142,27 @@ def _apply_rule(rule, record, parsed_record):
...
+                    record[attribute.key] = value
+        except Exception as ex:
+            logging.log_error_without_stacktrace(f"Encountered exception when evaluating attribute {attribute} of rule for {rule.entity_type_name}")
```
mmajcher, line 163: Same, where do we want to go with "as ex"? Keep it and log or remove

src/logs/metadata_engine/metadata_engine.py
mmajcher: OK I just found sorting :) Still not sure priority is correct word, it only touches ordering, does not say about this important thing that is allowing referencing in other places.

src/logs/metadata_engine/metadata_engine.py
```diff
@@ -194,13 +210,16 @@ def _create_attributes(attributes_json: List[Dict]) -> List[Attribute]:
...
         pattern = source_json.get("pattern", None)
 
         if key and pattern:
-            result.append(Attribute(key, pattern))
+            result.append(Attribute(key, priority, pattern))
         else:
             logging.warning(f"Encountered invalid rule attribute with missing parameter, parameters were: key = {key}, pattern = {pattern}")
 
+    # attributes without priority are executed last
+    result.sort(key= lambda attribute: attribute.priority if attribute.priority is not None else inf)
```
mmajcher, line 222: You know what. I think we are sorting entries in static json lots of times. Probably there is a readability reason for some specific ordering but maybe we should consider sorting it once and adjusting static json so we don't have to sort it in runtime anymore

src/logs/transformation.py
```diff
@@ -43,7 +43,7 @@ def extract_dt_logs_from_single_record(
         log_entry = transform_single_log_entry(log_event, batch_metadata, record_metadata)
         logs.append(log_entry)
 
-    log_content_lens = [len(log["content"]) for log in logs]
+    log_content_lens = [len(  ["content"] )   for log in logs    if log.get("content", None) is not None ]
```
mmajcher, line 46: This line looks weird, len with spaces and square brackets only?

src/config/cloudtrail.json
```diff
@@ -0,0 +1,1006 @@
...
+          "description": "If a field errorCode exists, then return Failure.{errorCode}, else return Succeeded",
+          "pattern": "if( log_content_parsed.errorCode==null, &'Succeeded', &[`Failed`, log_content_parsed.errorCode] | [? @!=null ] | join('.', @) , @)"
+        },
+        {
+          "key": "severity",
+          "pattern": "if( log_content_parsed.errorCode==null, &'INFO', &'ERROR', @)"
+        },
+        {
+          "key": "dt.source_entity",
+          "pattern": "if( log_content_parsed.eventName == 'AddRoleToDBCluster', &dt_meid_supporting_service_v2( 'Aurora', \"aws.arn\"), &EmptyValue, @ )"
```
mmajcher, line 40: Why escape double-quote when you can use single quote?

src/config/cloudtrail.json
```diff
@@ -0,0 +1,1006 @@
...
+          "pattern": "if( log_content_parsed.eventName == 'CreateDBSnapshot', &dt_meid_rds_v2( \"aws.arn\"), &EmptyValue, @ )"
+        },
+        {
+          "key": "aws.arn",
+          "priority": 10,
+          "pattern": "if( log_content_parsed.eventName == 'CreateDBSnapshot', &format_arn('arn:{}:rds:{}:{}:db:{}', [partition, region, account_id, log_content_parsed.requestParameters.dBInstanceIdentifier]), &EmptyValue, @ )"
+        },
+        {
+          "key": "aws.resource.id",
+          "pattern": "if( log_content_parsed.eventName == 'CreateDBSnapshot', &log_content_parsed.requestParameters.dBInstanceIdentifier, &EmptyValue, @ )"
```
mmajcher, line 212: Well, I don't know how much performance impact we have with those ifs. But they are clearly exclusive. Switch statement might reduce number of comparisons significantly, i.e. once you match one eventName (optimistically one in first half ;) ) then you don't need to compare rest of them. 
I am not sure If we want to change it anyhow, just pointing out that this is not optimal from performance point of view

src/config/cloudtrail.json
```diff
@@ -0,0 +1,1006 @@
...
+          "priority": 10,
+          "pattern": "if( log_content_parsed.eventName == 'CreateEventSubscription' && log_content_parsed.requestParameters.sourceType == 'db-cluster', &format_arn('arn:{}:rds:{}:{}:cluster:{}', [partition, region, account_id, log_content_parsed.requestParameters.sourceIds[0] ]), &EmptyValue, @ )"
+        },
+        {
+          "key": "aws.resource.id",
+          "pattern": "if( log_content_parsed.eventName == 'CreateEventSubscription' && log_content_parsed.requestParameters.sourceType == 'db-cluster', &log_content_parsed.requestParameters.sourceIds[0], &EmptyValue, @ )"
+        },
+        {
+          "key": "dt.source_entity",
+          "pattern": "if( log_content_parsed.eventName == 'CreateEventSubscription' && log_content_parsed.requestParameters.sourceType == 'db-instance', &dt_meid_rds_v2( \"aws.arn\"), &EmptyValue, @ )"
```
mmajcher, line 229: This if function does "visit" on branch expression. So we have those ampersand-deferred expressions, right? Can we put non-deferred expression, like a normal string without ampersand or just "null"?

mmajcher: Basically, can we have already evaluated "EmptyValue", so maybe put just "null" (without ampersand) so we don't have to evaluate &"EmptyValue" later?

src/config/cloudtrail.json
```diff
@@ -0,0 +1,1006 @@
...
+          "priority": 10,
+          "pattern": "if( contains(log_content_parsed.eventName, 'DeleteEventSubscription') && log_content_parsed.requestParameters.sourceType == 'db-cluster', &format_arn('arn:{}:rds:{}:{}:cluster:{}', [partition, region, account_id, log_content_parsed.responseElements.sourceIdsList[0] ]), &EmptyValue, @ )"
+        },
+        {
+          "key": "aws.resource.id",
+          "pattern": "if( contains(log_content_parsed.eventName, 'DeleteEventSubscription') && log_content_parsed.requestParameters.sourceType == 'db-cluster', &log_content_parsed.responseElements.sourceIdsList[0], &EmptyValue, @ )"
+        },
+        {
+          "key": "dt.source_entity",
+          "pattern": "if( contains(log_content_parsed.eventName, 'DeleteEventSubscription') && log_content_parsed.requestParameters.sourceType == 'db-instance', &dt_meid_rds_v2( \"aws.arn\"), &EmptyValue, @ )"
```
mmajcher, line 308: Yeah, same case here, lots of contains check that might be unnecessary in many cases.
Actually this might be optimised further if our switch makes a dictionary with keys of possible values ("DeleteEventSubscription") and then reference specific dictionary entry and evaluate this. 
I am not sure how big perf impact is but this metadata thing is run for each single entry so I think this is worth investigating.

mmajcher: I think this is actually a classic way for switches, e.g. compiled C does this, where it does not evaluate each condition but jumps straight to correct one with O(1). There are some assumptions to be held though

src/config/cloudtrail.json
```diff
@@ -0,0 +1,1006 @@
...
+          "pattern": "if( log_content_parsed.eventName == 'ModifyDBCluster', &dt_meid_supporting_service_v2( 'Aurora', \"aws.arn\"), &EmptyValue, @ )"
+        },
+        {
+          "key": "aws.arn",
+          "priority": 10,
+          "pattern": "if( log_content_parsed.eventName == 'ModifyDBCluster', &format_arn('arn:{}:rds:{}:{}:cluster:{}', [partition, region, account_id, log_content_parsed.requestParameters.dBClusterIdentifier]), &EmptyValue, @ )"
+        },
+        {
+          "key": "aws.resource.id",
+          "pattern": "if( log_content_parsed.eventName == 'ModifyDBCluster', &log_content_parsed.requestParameters.dBClusterIdentifier, &EmptyValue, @ )"
```
mmajcher, line 624: We do need a profiling run, as simple as possible but there are few places that might be perf heavy and I think we need just basic estimation which places have strong impact and which do not

mmajcher: But like really simple, 10000 x same cloudtrail log entry

src/config/lambda.json
```diff
@@ -24,15 +24,15 @@
...
         {
           "key": "aws.arn",
-          "pattern": "format('arn:{}:lambda:{}:{}:function:{}', [partition, region, account_id, function_name])"
+          "pattern": "format_arn('arn:{}:lambda:{}:{}:function:{}', [partition, region, account_id, function_name])"
         },
         {
-          "key": "dt.source_entity",
-          "pattern": "dt_meid_md5('AWS_LAMBDA_FUNCTION', format('{}{}_{}', [function_name, region, account_id]))"
+          "key": "severity",
+          "pattern": "if( starts_with(log_content_parsed, '[ERROR]'), &'ERROR', &'INFO', @)"
```
mmajcher, line 31: Are you parsing the log here? Is there anything to be parsed, I thougth lambda logs are a text, not json.

src/logs/metadata_engine/metadata_engine.py
```diff
@@ -139,14 +142,27 @@ def _apply_rule(rule, record, parsed_record):
     if rule.aws_loggroup_pattern and "log_group" in record:
         extracted_values = parse_aws_loggroup_with_grok_pattern(record["log_group"], rule.aws_loggroup_pattern)
         record.update(extracted_values)
+    if rule.log_content_parse_type == "json":
+        try:
+            record["log_content_parsed"] = json.loads(parsed_record.get("content", {}))
+        except Exception as ex:
+            logging.log_error_with_stacktrace(ex, f"Encountered exception when parsing log content as json, requested by rule for {rule.entity_type_name}")
+    else:
+        record["log_content_parsed"] = parsed_record.get("content", "")
```
mmajcher, line 151: Maybe then "content" should be "content" and not "log_content_parsed"?

src/logs/metadata_engine/metadata_engine.py
mmajcher: Ok, my mistake, this load&sort is only executed once per lambda run so not an issue 👍

src/logs/metadata_engine/jmespath.py
```diff
@@ -44,26 +44,96 @@ def _func_if(self, condition, if_true_expression, if_false_expression, node_scop
...
+
+    # aws credentials v1 - md5
+    @functions.signature({'types': ['string', 'null']})
+    def _func_dt_meid_elb_v1(self, dnsName): return me_id.meid_md5("ELASTIC_LOAD_BALANCER", dnsName)
+    @functions.signature({'types': ['string', 'null']})
```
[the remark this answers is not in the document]
wojtasskorcz: Agreed, a newline would make it more readable

src/logs/metadata_engine/jmespath.py
```diff
@@ -44,26 +44,96 @@ def _func_if(self, condition, if_true_expression, if_false_expression, node_scop
...
-
+            if value is not None:
+                output = output.replace("{}", value, 1)
         return output
 
-    @functions.signature({'types': ['string']},
-                         {'types': ['string']})
+    @functions.signature({'types': ['string', 'null']},
+                         {"types": ['array']})
+    def _func_format_arn(self, pattern, values):
```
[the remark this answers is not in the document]
wojtasskorcz: Agreed, this may come in handy in other places and will be easier to spot by a person needing it in the future if we name it more explicitly (for example, like Marek recommended)

src/logs/metadata_engine/jmespath.py
```diff
@@ -44,26 +44,96 @@ def _func_if(self, condition, if_true_expression, if_false_expression, node_scop
...
+        if None in values:
+            # any None value in input is treated as missing required data, no complete arn can be calculated
+            return None
+
+        return self.format(pattern, values)
+
+    @functions.signature({'types': ['string', 'null']},
+                         {'types': ['string', 'null']})
     def _func_dt_meid_md5(self, entity_type, hashing_input):
```
wojtasskorcz, line 80: Is it used anymore now that we have dedicated functions for each type?

src/logs/metadata_engine/jmespath.py
wojtasskorcz: Agreed that it's minor but I also favor the `format("{}{}_{}", functionName, region, accountId) notation. It also clearly separates the first parameter (`entity_type`) from `hashing_input`. Here those two parts are "mixed together"

src/logs/metadata_engine/metadata_engine.py
wojtasskorcz: If you look in the implementation of `logging.exception()`, the exception details are retrieved using `traceback`: https://docs.python.org/3/library/traceback.html

mmajcher: Sounds good, but then why this "as ex" part? I think it is not needed in such case

src/logs/metadata_engine/metadata_engine.py
wojtasskorcz: Yeah, let's not implement throttling here. This PR is huge anyway :)

src/config/cloudtrail.json
wojtasskorcz: Sound like a good optimization, then we'd only need a few additional if-comparisons. As this would be a pretty significant change to the JSON schema, I'd do it after the simple profiling you proposed in another comment to check if it's worth the effort. WDYT @dt-be ?

src/config/cloudtrail.json
mmajcher: And if we raise this switch-dictionary to schema (new field: "conditionalPattern" with "sourceField" and a map value->expression) then this dictionary is built only once when loading configs

src/logs/metadata_engine/metadata_engine.py
dt-be: why not, now that I have the else for lambda. Will call it "log_content", though.

src/config/cloudtrail.json
dt-be: `Can we put non-deferred expression` no, it currently requires deferred, I'm not sure I want to fight to change that

 I had some trouble with plain null, that's why I ended with EmptyValue. Might recheck again if I will remember.

src/config/cloudtrail.json
dt-be: Eeee, this contains should be gone, it's a left-over from an attempt to have reduce the number of ifs by using *.

src/config/cloudtrail.json
dt-be: I have a setup with some thousand entries, but haven't run it yet - I plan to do it tomorrow

src/config/cloudtrail.json
dt-be: There are a couple pretty invasive things that could be done here. I would like to go in any of those directions only based on profiling results.

src/logs/metadata_engine/metadata_engine.py
mmajcher: If it's helpful for you then it's cool, I thought it's some leftover

src/logs/metadata_engine/me_id.py
mmajcher: At least put that hex from java in a comment, for reference

src/logs/metadata_engine/metadata_engine.py
mmajcher: Nah, let's leave it, I think now it's not that important

src/config/cloudtrail.json
mmajcher: No, if it does require deferred then I also see no point, let's leave this

src/config/lambda.json
mmajcher: I was really looking at this "parsed" thing in the name, it is addressed in other comment. I do understand the need to do this starts_with check 👍

src/logs/metadata_engine/jmespath.py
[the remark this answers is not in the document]
mmajcher: I understand your reasoning, it's all good! I just feel like trying to convey such additional meaning through function name gets a little confusing. If I see arn/id field in config with function "format_required" or "format_require_nonnulls" or whatever then I might realise this reasoning easier

src/logs/metadata_engine/me_id.py
dt-be: Yes, that's what I did + a jira for reference
https://github.com/dynatrace-oss/dynatrace-aws-log-forwarder/pull/15/commits/4af3dc0188dc1f257d8f5fbf636daa10069152bb

src/config/cloudtrail.json
dt-be: First profiling result: 
I ran in first on a much smaller scale, with real data from my AWS setup 70-90 log entries that was still running 20-30 sec. I managed to get it down to 2-3s.

The story was similar to that with Grok - initialization happening too often. The default limit for number of cached/"compiled" Jmespath patterns/expressions was 128. With CloudTrail itself having 228 entries, this resulted in jmespath.parser.parse() reparsing all the patterns on for every logEntry and taking most of the execution time.

So this had nothing to do with if/switch of ==/contains ;) 

I will still do some further tests.

mmajcher: It seems it is still a lot (2-3 sek for 70-90 entries)

dt-be: I've run setups with different numbers of attribute patterns, most of them resulting in execution duration exceeding Lambda timeout of 60s (both in AWS and local setup):

15k, max_size=20000 > 60s
7,5k > 55-60s
5k > 50-60s

I'm setting the cache limit deliberately high, as it seems to have no noticable memory penalty in AWS setup, whereas the penalty for not having them cached was severe. While it barely works at my test of 5k in a single config file, I still would rather leave space for any improvements elsewhere + a scenario of it being thousands of attribute patterns but in various files - rather than risk recompiling the patterns per each log entry again. 

But we can discuss how high the limit should be - I'm setting it at 20k attribute patterns for now.

mmajcher:
I think we also need to update the docs with service-specific docs. 
For sure we need info that you need to name log group for RDS cloudtrail in specific way. 
But also probably we should add a table with a list of supported services that says where we have linking and service recognition? @wojtasskorcz

src/config/cloudtrail.json
mmajcher: Given current number of patterns, can we ingest 10MB RDS cloudtrail logs per minute?

dt-be: No, we can't - as discussed on a meeting, we're considering it too slippery for now to go along with it. 

I'm pushing current changes to a separate branch: APM-299389-metadata-extraction-and-entity-mapping-for-cloudtrail-rds_declined
In this PR branch I'm reverting from cloudtrail config all entity mapping (leaving only the most generic CloudTrail fields that aren't event-specific) and the tests section for those. All the rest stays the same.

mmajcher (approved):

wojtasskorcz:
Don't now why I can't reply directly to @mmajcher 's comment above but I agree that it would be good to include info about which services we support in the official documentation. Let's have it as a separate ticket as it's not directly related to these changes

wojtasskorcz (approved):

PR #17, opened 2021-06-17
Feature/apm 297893 deploy ag

mmajcher:


dynatrace-aws-logs.sh
```diff
@@ -127,9 +143,22 @@ arguments:
...
   if [ -z "$REQUIRE_VALID_CERTIFICATE" ]; then REQUIRE_VALID_CERTIFICATE="false"; fi
   if [ -z "$STACK_NAME" ]; then STACK_NAME=$DEFAULT_STACK_NAME; fi
+  if [ -z "$USE_EXISTING_ACTIVE_GATE" ]; then echo "No --use-existing-active-gate"; print_help_deploy; exit 1; fi
 
   if [[ "$REQUIRE_VALID_CERTIFICATE" != "true" ]] && [[ "$REQUIRE_VALID_CERTIFICATE" != "false" ]];
     then echo "Invalid value for parameter --require-valid-certificate. Provide 'true' or 'false'"; print_help_deploy; exit 1; fi
+  if [[ "$USE_EXISTING_ACTIVE_GATE" != "true" ]] && [[ "$USE_EXISTING_ACTIVE_GATE" != "false" ]];
+    then echo "Invalid value for parameter --use-existing-active-gate. Provide 'true' or 'false'"; print_help_deploy; exit 1; fi
+  if [[ "$USE_EXISTING_ACTIVE_GATE" == "false" ]]; then
+    if [ -z "$PAAS_TOKEN" ]; then echo "No --paas-token (required when --use-existing-active-gate=false)"; print_help_deploy; exit 1; fi
```
dt-be, line 153: 1. minor, I'd merge if 152 & 153
2. For me this message is a bit confusing (existing=false) doesn't convey real reason why in this case it is required.  Not sure if it's way better, but something in the direction of: 
`--paas-token is required to allow installation a new ActiveGate. To use an existing one change --use-existing-active-gate=true`
But this sounds already more like part of a doc then command line message .So maybe just "no --paas-token" like with the other messages and then help is printed either way.

dynatrace-aws-log-forwarder-template.yaml
```diff
@@ -518,6 +563,128 @@ Resources:
         - __Region__: !Ref 'AWS::Region'
           __LambdaName__: !Ref Lambda
 
+  EC2ActiveGate:
+    Condition: DeployVPCwithAG
+    Type: AWS::EC2::Instance
+    DependsOn: VPCPublicRouteInternet
+    Properties:
+      ImageId: !Ref LatestAmazonLinuxAmiId
+      InstanceType: "t3.small"
```
wojtasskorcz, line 572: Is it enough for production use? Or was it only used for your tests?

dt-be: I was asking myself the same question. 
I was also wondering if we allow the customer to explicitly choose different/bigger machine size if they don't like the default. But I got stuck on checking what Azure has.

dynatrace-aws-log-forwarder-template.yaml
```diff
@@ -518,6 +563,128 @@ Resources:
...
+            DeleteOnTermination: "true"
+            VolumeSize: "8"
+      SecurityGroupIds: [ !Ref VPCSecurityGroup ]
+      UserData:
+        Fn::Base64:
+          !Sub
+          - |
+            #!/bin/bash -xe
+            wget -O Dynatrace-ActiveGate-Linux-x86.sh "${DynatraceEnvironmentUrl}/api/v1/deployment/installer/gateway/unix/latest?arch=x86&flavor=default" --header="Authorization: Api-Token ${DynatracePaasToken}"
+            wget https://ca.dynatrace.com/dt-root.cert.pem ; ( echo 'Content-Type: multipart/signed; protocol="application/x-pkcs7-signature"; micalg="sha-256"; boundary="--SIGNED-INSTALLER"'; echo ; echo ; echo '----SIGNED-INSTALLER' ; cat Dynatrace-ActiveGate-Linux-x86.sh ) | openssl cms -verify -CAfile dt-root.cert.pem > /dev/null
```
dt-be, line 587: "-O name" just in case or not worth it?

dynatrace-aws-log-forwarder-template.yaml
mmajcher: image


This one has 2vCPU and 2GB ram.
You can't install AG with less than 2GB ram. You can start AG in container, though.

dynatrace-aws-log-forwarder-template.yaml
mmajcher: Not worth it, this instance is as clean as it gets :D

dynatrace-aws-logs.sh
mmajcher: 1. Right, will fix 
2. So basically just remove this explanation in parens? Sure. Yes, it is more explained in help and help is printed in this case

dynatrace-aws-log-forwarder-template.yaml
mmajcher: I did not test it with large workloads but would assume it is a production size, in AG docs 2GB is minimum memory. (4GB is recommended but our AG only pushes logs, all other modules are disabled, so not very overworked and should be fine)

dynatrace-aws-log-forwarder-template.yaml
```diff
@@ -518,6 +563,128 @@ Resources:
...
+            echo "[beacon_forwarder]" >> /var/lib/dynatrace/gateway/config/custom.properties
+            echo "beacon_forwarder_enabled = false" >> /var/lib/dynatrace/gateway/config/custom.properties
+            echo "[extension_controller]" >> /var/lib/dynatrace/gateway/config/custom.properties
+            echo "extension_controller_enabled = false" >> /var/lib/dynatrace/gateway/config/custom.properties
+            echo "[dbAgent]" >> /var/lib/dynatrace/gateway/config/custom.properties
+            echo "dbAgent_enabled = false" >> /var/lib/dynatrace/gateway/config/custom.properties
+            echo "[metrics_ingest]" >> /var/lib/dynatrace/gateway/config/custom.properties
+            echo "metrics_ingest_enabled = false" >> /var/lib/dynatrace/gateway/config/custom.properties
+            echo "[collector]" >> /var/lib/dynatrace/gateway/config/custom.properties
+            echo "MSGrouter = false" >> /var/lib/dynatrace/gateway/config/custom.properties
```
dt-be, line 611: I thought maybe something like this would be clearer, unless there's some reason against it:

```
echo "[aws_monitoring] 
[azure_monitoring] 
azure_monitoring_enabled = false 
[cloudfoundry_monitoring] 
cloudfoundry_monitoring_enabled = false 
[kubernetes_monitoring] 
kubernetes_monitoring_enabled = false 
[vmware_monitoring] 
vmware_monitoring_enabled = false 
[rpm] 
rpm_enabled = false 
[beacon_forwarder] 
beacon_forwarder_enabled = false 
[extension_controller] 
extension_controller_enabled = false 
[dbAgent] 
dbAgent_enabled = false 
[metrics_ingest] 
metrics_ingest_enabled = false 
[collector] 
MSGrouter = false" | cat > /var/lib/dynatrace/gateway/config/custom.properties
```

dynatrace-aws-log-forwarder-template.yaml
dt-be: Question, does this verification actually change/stop the flow? I guess it the one in the usual Deploy OneAgent page, and perhaps it can stay for debugging either way, just wondering.

mmajcher: It should return non zero code and this will stop execution (-e bash option above). I will double check

dynatrace-aws-log-forwarder-template.yaml
wojtasskorcz: I think I like it better too

dynatrace-aws-log-forwarder-template.yaml
mmajcher: It returns non zero code (4) and this will stop the execution

dynatrace-aws-log-forwarder-template.yaml
mmajcher: Nah, looks good, just need to check if this is working correcty regarding indentation

dynatrace-aws-log-forwarder-template.yaml
```diff
@@ -30,6 +30,27 @@ Parameters:
...
+    Default: "not_provided"
+    Description: "Needed to construct full target-url path in scenario with new AG deployment (with existing AG option, target-url is already provided by DynatraceEnvironmentUrl)"
+  LatestAmazonLinuxAmiId:
+    # this will query for latest AMI ID automatically
+    # https://aws.amazon.com/blogs/compute/query-for-the-latest-amazon-linux-ami-ids-using-aws-systems-manager-parameter-store/
+    Type: 'AWS::SSM::Parameter::Value<AWS::EC2::Image::Id>'
+    Default: '/aws/service/ami-amazon-linux-latest/amzn2-ami-hvm-x86_64-gp2'
+
+Conditions:
+  DeployVPCwithAG: !Equals [ !Ref UseExistingActiveGate, 'false' ]
```
dt-be, line 53: imho it's more DeployAGwithVPC, the way it is makes it bit harder to read the conditions later on

dynatrace-aws-logs.sh
```diff
@@ -139,6 +167,7 @@ arguments:
 
   aws cloudformation deploy --stack "$STACK_NAME" --template-file "$TEMPLATE_FILE" --capabilities CAPABILITY_IAM \
     --parameter-overrides DynatraceEnvironmentUrl="$TARGET_URL" DynatraceApiKey="$TARGET_API_TOKEN" VerifySSLTargetActiveGate="$REQUIRE_VALID_CERTIFICATE" \
+    UseExistingActiveGate="$USE_EXISTING_ACTIVE_GATE" TenantIdNewAGonly="$TENANT_ID" DynatracePaasToken="$TARGET_PAAS_TOKEN" \
```
dt-be, line 170: I'd just call it TenantId, it's bit noisy to read in cloudFormation later on the way it is

dynatrace-aws-logs.sh
```diff
@@ -53,14 +53,18 @@ case $MAIN_OPTION in
...
     printf "
-usage: dynatrace-aws-logs.sh deploy --target-url TARGET_URL --target-api-token TARGET_API_TOKEN [--require-valid-certificate {true|false}] [--stack-name STACK_NAME]
+usage: dynatrace-aws-logs.sh deploy --target-url TARGET_URL --target-api-token TARGET_API_TOKEN --use-existing-active-gate {true|false} [--target-paas-token TARGET_PAAS_TOKEN] [--require-valid-certificate {true|false}] [--stack-name STACK_NAME]
 
 arguments:
   -h, --help            show this help message and exit
   --target-url TARGET_URL
-                        The URL to Your Dynatrace SaaS environment logs ingest target. Currently only ActiveGate endpoint is supported: https://<active_gate_address>:9999/e/<environment_id> (e.g. https://22.111.98.222:9999/e/abc12345)
+                        The URL to Your Dynatrace SaaS environment logs ingest target. If you choose new ActiveGate deployment (--use-existing-active-gate=false), provide Tenant URL (https://<your_environment_ID>.live.dynatrace.com). If you choose to use existing ActiveGate (--use-existing-active-gate=true), provide ActiveGate endpoint: https://<active_gate_address>:9999/e/<environment_id> (e.g. https://22.111.98.222:9999/e/abc12345).
```
dt-be, line 61: I like Azure docs wording (https://www.dynatrace.com/support/help/shortlink/azure-log-fwd#dynatrace-) + I'd go with true first, rather then false, so maybe something in this direction:
```
If you want to use an existing ActiveGate (--use-existing-active-gate=true), set it to your ActiveGate endpoint: https://<your_activegate_IP_or_hostname>:9999/e/<your_environment_ID> If you want to deploy new ActiveGate in AWS EC2 (--use-existing-active-gate=true), set it to your Dynatrace SaaS environment: https://<your_environment_ID>.live.dynatrace.com
```

dynatrace-aws-log-forwarder-template.yaml
```diff
@@ -30,6 +30,27 @@ Parameters:
...
+      - 'true'
+      - 'false'
+  DynatracePaasToken:
+    Description: Dynatrace PaaS token
+    Type: String
+    NoEcho: true
+  TenantIdNewAGonly:
+    Type: String
+    Default: "not_provided"
+    Description: "Needed to construct full target-url path in scenario with new AG deployment (with existing AG option, target-url is already provided by DynatraceEnvironmentUrl)"
```
dt-be, line 45: I got lost, so just to make sure I got this right:
`--use-existing-ag=true --target-url=AG_URL`
< works as it used to

`use-existing-ag=false --target-url=TENANT_URL`
< extract TenantId / TenantIdNewAGonly
< TenantId to calculate AG endpoint: https://${EC2ActiveGateIP}:9999/e/${TenantId}"

If yes then trying to think of something to make it clearer, not sure if that would do, or is still the same:
`Needed in scenario with new AG deployment to calculate an AG endpoint: https://${EC2ActiveGateIP}:9999/e/${TenantId} for the proper tenant`

dynatrace-aws-log-forwarder-template.yaml
mmajcher: Maybe DeployAGandVPC? Then in future we can add DeployAGintoExistingVPC?

dynatrace-aws-logs.sh
mmajcher: True that, changing

dynatrace-aws-logs.sh
mmajcher: Check out the docs
https://bitbucket.lab.dynatrace.org/projects/WX/repos/doc/pull-requests/6090/overview
We are going with something like this 
image
So we kinda push the option with new AG to the top, as it's the easy way to do it (targeting non advanced user that might not have private AG). 

So we can either
1) Discuss ordering in docs and possibly change if you see it other way
2) Change this wording here and have this little order mixing
How do you see it?

mmajcher: I try to be consistent with Azure; personally, I would prefer different params altogether, like 
'--tenant-url' if you want AG or '--active-gate-url' if you want to use existing 
and get rid of this --use-existing-active-gate flag which is like "main controller" for params right now

dynatrace-aws-log-forwarder-template.yaml
mmajcher: I am not sure I understand this notation of yours 100% but I think you are correct.

Why this `for the proper tenant`? This part I do not understand also

mmajcher: If you go with newAG then you need to provide TenantId (bash script extracts this from TenantUrl)

mmajcher: Maybe this last part in parens is confusing? I think it might be looking at it right now

dynatrace-aws-log-forwarder-template.yaml
dt-be: My thinking was more along these lines
DeployAGwithVPC > to me sounds like deploy AG and deploy VPC
DeployAGtoExistingVPC

but DeployAGandVPC is already good

dynatrace-aws-logs.sh
dt-be: True, I also just noticed in the docs PR that new AG goes first. Personally, different order in different places makes sense to me, as I don't expect somebody going from both doc and usage at the same time. But we can go with doc order.

I won't say consistency doesn't hurt me here a bit, too, but let's have it same as Azure;)

But I'm thinking, since this is formatted as is, can't we make `If true` `If false` for both `--target-url` and `--use-existing-active-gate` and go begin each from a new line, rather then in a single line? I don't mind the ordering at all then.

dynatrace-aws-log-forwarder-template.yaml
dt-be: Yeap, I'd get rid of this part: `(with existing AG option, target-url is already provided by DynatraceEnvironmentUrl)`

> Why this for the proper tenant? This part I do not understand also

I meant that we would know AG address itself (https://${EC2ActiveGateIP}:9999), but we need also the path for the tenantId on this AG (https://${EC2ActiveGateIP}:9999**/e/${TenantId}** for the proper tenant)

dynatrace-aws-log-forwarder-template.yaml
mmajcher: Works fine 👍 updated

dynatrace-aws-log-forwarder-template.yaml
mmajcher: Updated to DeployAGwithVPC

dynatrace-aws-log-forwarder-template.yaml
mmajcher: Rewritten this description, please check if it is good enough now

dynatrace-aws-logs.sh
mmajcher: I just updated this, please check. I think it is more readable now with clean separated cases - each in new line

dt-be (dismissed):

wojtasskorcz (approved):

kamiljsokolowski (dismissed):
As most of my concerns were already addressed by other reviewers... approved.

dt-be (approved):

wojtasskorcz (approved):

PR #18, opened 2021-06-28
APM-280447

kamiljsokolowski:
This update will allow the creation and deletion of resources required for the e2e test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ notifications:
     on_success: never
 stages:
 - tests
+- build
 - deployment
 before_install:
   # set up awscli packages
@@ -24,17 +25,24 @@ jobs:
         - docker build -f pipeline/Dockerfile -t lambda_image .
         - docker run --rm lambda_image bash -c 'python3 -m pylint --rcfile=pipeline/pylint.cfg $(find src/ -name "*.py")'
         - docker run --rm lambda_image bash -c 'cd src && python3 -m pytest -v ../tests/unit'
+    - stage: build
+      name: Build AWS Lambda package
+      #if: tag =~ /^release.*$/
+      language: shell
+      services: docker
+      script:
+        - docker run --rm -v `pwd`:/var/task amazon/aws-sam-cli-build-image-python3.8 bash -c './build-release-package.sh'
+        - cp -p dynatrace-aws-log-forwarder.zip ~/build-stages-shared-storage/$TRAVIS_BUILD_NUMBER
     - stage: deployment
       name: Github Release Deployment
       if: tag =~ /^release.*$/
       language: shell
       services: docker
-      script:
-        - docker run --rm -v `pwd`:/var/task amazon/aws-sam-cli-build-image-python3.8 bash -c './build-release-package.sh'
+      script: skip
       deploy:
         provider: releases
         api_key: $GITHUB_RELEASE_API_KEY
-        file: ./dynatrace-aws-log-forwarder.zip
+        file: ~/build-stages-shared-storage/$TRAVIS_BUILD_NUMBER/dynatrace-aws-log-forwarder.zip
         skip_cleanup: true
         on:
           tags: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ jobs:
       services: docker
       before_script: cd ~/build-stages-shared-storage/$TRAVIS_BUILD_NUMBER
       script:
-        - unzip dynatrace-aws-log-forwarder-lambda.zip
+        - unzip dynatrace-aws-log-forwarder.zip
         # deploy log forwarder resources reqiured for e2e to AWS
         - ./dynatrace-aws-logs.sh deploy --target-url $DYNATRACE_ENV_URL --target-api-token $DYNATRACE_API_KEY --use-existing-active-gate true --require-valid-certificate $VERIFY_SSL --stack-name aws-log-forwarder-e2e-build-$TRAVIS_BUILD_NUMBER
         # delete log forwarder AWS resources
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,7 @@ jobs:
         - ./dynatrace-aws-logs.sh deploy \
           --target-url $DYNATRACE_ENV_URL \
           --target-api-token $DYNATRACE_API_KEY \
-          --require-valid-certificate $VERIFY_SSL
+          --require-valid-certificate $VERIFY_SSL \
           --stack-name aws-log-forwarder-e2e-build-$TRAVIS_BUILD_NUMBER
         # delete log forwarder AWS resources
         - ./dynatrace-aws-logs.sh delete \
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,7 @@
+env:
+  global:
+    # include $HOME/.local/bin for `aws`
+    - PATH=$HOME/.local/bin:$PATH
 language: shell
 notifications:
   email:
@@ -5,6 +9,11 @@ notifications:
 stages:
 - tests
 - deployment
+before_install:
+  # set up awscli packages
+  - pip install --user awscli
+  - mkdir -p ~/build-stages-shared-storage/$TRAVIS_BUILD_NUMBER
+  - aws s3 sync s3://aws-log-forwader-travis-build-stages-shared-storage/$TRAVIS_BUILD_NUMBER ~/build-stages-shared-storage/$TRAVIS_BUILD_NUMBER
 jobs:
   include:
     - stage: tests
@@ -29,3 +38,6 @@ jobs:
         skip_cleanup: true
         on:
           tags: true
+
+after_success:
+  - aws s3 sync s3://aws-log-forwader-travis-build-stages-shared-storage/$TRAVIS_BUILD_NUMBER
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,9 @@ jobs:
           --target-api-token $DYNATRACE_API_KEY \
           --require-valid-certificate $VERIFY_SSL
           --stack-name aws-log-forwarder-e2e-build-$TRAVIS_BUILD_NUMBER
+        # delete log forwarder AWS resources
+        - ./dynatrace-aws-logs.sh delete \
+          --stack-name aws-log-forwarder-e2e-build-$TRAVIS_BUILD_NUMBER
     - stage: deployment
       name: Github Release Deployment
       if: tag =~ /^release.*$/
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ before_install:
   - pip install --user awscli
   # set up shared storage for build artifacts
   - mkdir -p ~/build-stages-shared-storage/$TRAVIS_BUILD_NUMBER
-  - aws s3 sync s3://aws-log-forwader-travis-build-shared-storage/$TRAVIS_BUILD_NUMBER ~/build-stages-shared-storage/$TRAVIS_BUILD_NUMBER
+  - aws s3 sync s3://aws-log-forwarder-travis-build-shared-storage/$TRAVIS_BUILD_NUMBER ~/build-stages-shared-storage/$TRAVIS_BUILD_NUMBER
 jobs:
   include:
     - stage: unit_tests
@@ -62,4 +62,4 @@ jobs:
           tags: true
 
 after_success:
-  - aws s3 sync ~/build-stages-shared-storage/$TRAVIS_BUILD_NUMBER s3://aws-log-forwader-travis-build-shared-storage/$TRAVIS_BUILD_NUMBER
+  - aws s3 sync ~/build-stages-shared-storage/$TRAVIS_BUILD_NUMBER s3://aws-log-forwarder-travis-build-shared-storage/$TRAVIS_BUILD_NUMBER
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,6 +46,7 @@ jobs:
         - cp -r ~/build-stages-shared-storage/$TRAVIS_BUILD_NUMBER/dynatrace-aws-log-forwarder-lambda.zip .
         - ./dynatrace-aws-logs.sh deploy --target-url $DYNATRACE_ENV_URL --target-api-token $DYNATRACE_API_KEY --use-existing-active-gate true --require-valid-certificate $VERIFY_SSL --stack-name aws-log-forwarder-e2e-build-$TRAVIS_BUILD_NUMBER
         # delete log forwarder AWS resources
+        - aws cloudformation delete-stack --stack-name aws-log-forwarder-e2e-build-$TRAVIS_BUILD_NUMBER
     - stage: deployment
       name: Github Release Deployment
       if: tag =~ /^release.*$/
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,7 @@ jobs:
       script:
         - docker run --rm -v `pwd`:/var/task amazon/aws-sam-cli-build-image-python3.8 bash -c './build-release-package.sh'
         - cp -p dynatrace-aws-log-forwarder.zip ~/build-stages-shared-storage/$TRAVIS_BUILD_NUMBER
+        - cp -p dynatrace-aws-log-forwarder-lambda.zip ~/build-stages-shared-storage/$TRAVIS_BUILD_NUMBER
     - stage: e2e_tests
       name: Execute e2e tests
       # run e2e tests only for development branches until test execution part is done
@@ -44,7 +45,6 @@ jobs:
       script:
         # deploy log forwarder resources reqiured for e2e to AWS
         - cp -r ~/build-stages-shared-storage/$TRAVIS_BUILD_NUMBER/dynatrace-aws-log-forwarder.zip .
-        - unzip dynatrace-aws-log-forwarder.zip
         - ./dynatrace-aws-logs.sh deploy --target-url $DYNATRACE_ENV_URL --target-api-token $DYNATRACE_API_KEY --require-valid-certificate $VERIFY_SSL --stack-name aws-log-forwarder-e2e-build-$TRAVIS_BUILD_NUMBER
         # delete log forwarder AWS resources
         - ./dynatrace-aws-logs.sh delete --stack-name aws-log-forwarder-e2e-build-$TRAVIS_BUILD_NUMBER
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,31 +1,66 @@
...
+      #if: tag =~ /^release.*$/
+      language: shell
+      services: docker
+      script:
+        - docker run --rm -v `pwd`:/var/task amazon/aws-sam-cli-build-image-python3.8 bash -c './build-release-package.sh'
+        - cp -p dynatrace-aws-log-forwarder.zip ~/build-stages-shared-storage/$TRAVIS_BUILD_NUMBER
+        - cp -p dynatrace-aws-log-forwarder-lambda.zip ~/build-stages-shared-storage/$TRAVIS_BUILD_NUMBER
+    - stage: e2e_tests
+      name: Execute e2e tests
+      # run e2e tests only for development branches until test execution part is done
```
sq2gxo, line 41: why? I see value in running deployment test without final test ready

.travis.yml
```diff
@@ -1,31 +1,66 @@
...
+      if: branch IN (APM-280447,APM-306464)
+      language: shell
+      services: docker
+      script:
+        # deploy log forwarder resources reqiured for e2e to AWS
+        - cp -r ~/build-stages-shared-storage/$TRAVIS_BUILD_NUMBER/dynatrace-aws-log-forwarder-lambda.zip .
+        - ./dynatrace-aws-logs.sh deploy --target-url $DYNATRACE_ENV_URL --target-api-token $DYNATRACE_API_KEY --require-valid-certificate $VERIFY_SSL --stack-name aws-log-forwarder-e2e-build-$TRAVIS_BUILD_NUMBER
```
sq2gxo, line 48: where are those ($DYNATRACE_*) ENV vars defined?

dynatrace-aws-logs.sh
```diff
@@ -150,6 +151,69 @@ arguments:
...
+        ensure_param_value_given $1 $2
+        STACK_NAME=$2
+        shift;shift;
+      ;;
+
+      "-h")
+        print_help_delete
+        shift; exit 0
+      ;;
+      "--help")
```
sq2gxo, line 193: tip - you can use `"-h" | "--help")` in case to not duplicate code

.travis.yml
kamiljsokolowski: There is value in running deployment without actually testing the log forwarder functionality. My motivation here was to save resources and build time until the entire suite is completed. 

Point taken. I will update.

.travis.yml
kamiljsokolowski: Wojtek mentioned, that you can define env variables using TravisCI GUI (`repo/more options/settings`).

dynatrace-aws-logs.sh
kamiljsokolowski: Thanks. Will update.

.travis.yml
kamiljsokolowski: Code updated.

dynatrace-aws-logs.sh
kamiljsokolowski: Done.

dynatrace-aws-logs.sh
```diff
@@ -150,6 +147,65 @@ arguments:
   aws cloudformation describe-stacks --stack-name "$STACK_NAME" --query "Stacks[0].Outputs"
   ;;
 
+"delete")
+
+  function print_help_delete {
```
mmajcher, line 152: Are we supposed to add this function? 
If we add this, I would expect this does delete stack AND removes all log group subscriptions for this stack. 
Also, this probably needs corresponding description in DT docs

dynatrace-aws-logs.sh
```diff
@@ -150,6 +147,65 @@ arguments:
...
+    printf "
+usage: dynatrace-aws-logs.sh delete [--stack-name STACK_NAME]
+
+arguments:
+  -h, --help            show this help message and exit
+  --stack-name STACK_NAME
+                        Optional. The name for the CloudFormation stack in which the resources will be deployed. This defaults to \"$DEFAULT_STACK_NAME\"
```
mmajcher, line 159: "will be deployed" not correct in this option :D

.travis.yml
```diff
@@ -1,31 +1,65 @@
...
+        - cp -p dynatrace-aws-log-forwarder-lambda.zip ~/build-stages-shared-storage/$TRAVIS_BUILD_NUMBER
+    - stage: e2e_tests
+      name: Execute e2e tests
+      if: branch = master OR type = pull_request
+      language: shell
+      services: docker
+      script:
+        # deploy log forwarder resources reqiured for e2e to AWS
+        - cp -r ~/build-stages-shared-storage/$TRAVIS_BUILD_NUMBER/dynatrace-aws-log-forwarder-lambda.zip .
+        - ./dynatrace-aws-logs.sh deploy --target-url $DYNATRACE_ENV_URL --target-api-token $DYNATRACE_API_KEY --require-valid-certificate $VERIFY_SSL --stack-name aws-log-forwarder-e2e-build-$TRAVIS_BUILD_NUMBER
```
mmajcher, line 47: Please confirm if I understand correctly. 
You are using script from repository? But with lambda.zip from previous step. If it is so, it does not test the full release zip package.

I would expect: you take release zip from build step. You copy it to new dir. Unzip in new dir. Perform deployment in there so we are sure you are not taking things from repo (but from zip as expected).
The artifact that we release it this full release zip and e2e should only be using this artifact and nothing from other places

.travis.yml
```diff
@@ -1,31 +1,65 @@
...
         - docker run --rm lambda_image bash -c 'cd src && python3 -m pytest -v ../tests/unit'
+    - stage: build
+      name: Build AWS Lambda package
+      #if: tag =~ /^release.*$/
+      language: shell
+      services: docker
+      script:
+        - docker run --rm -v `pwd`:/var/task amazon/aws-sam-cli-build-image-python3.8 bash -c './build-release-package.sh'
+        - cp -p dynatrace-aws-log-forwarder.zip ~/build-stages-shared-storage/$TRAVIS_BUILD_NUMBER
+        - cp -p dynatrace-aws-log-forwarder-lambda.zip ~/build-stages-shared-storage/$TRAVIS_BUILD_NUMBER
```
mmajcher, line 38: I think lambda.zip is included in main release zip (line above)? Not sure if I identify this zip above correctly

.travis.yml
```diff
@@ -1,31 +1,65 @@
...
         provider: releases
         api_key: $GITHUB_RELEASE_API_KEY
-        file: ./dynatrace-aws-log-forwarder.zip
+        file: ~/build-stages-shared-storage/$TRAVIS_BUILD_NUMBER/dynatrace-aws-log-forwarder.zip
         skip_cleanup: true
         on:
           tags: true
+
+after_success:
+  - aws s3 sync ~/build-stages-shared-storage/$TRAVIS_BUILD_NUMBER s3://aws-log-forwader-travis-build-shared-storage/$TRAVIS_BUILD_NUMBER
```
mmajcher, line 65: Type in s3 "forwader"

dynatrace-aws-logs.sh
```diff
@@ -22,7 +22,7 @@ function print_help_main_options {
   echo ""
   printf \
 "usage: dynatrace-aws-logs.sh [-h]
-                              {deploy,subscribe,unsubscribe,discover-log-groups}
+                              {deploy,delete,subscribe,unsubscribe,discover-log-groups}
```
mmajcher, line 25: I think delete is more like last option :)

dynatrace-aws-logs.sh
```diff
@@ -34,6 +34,7 @@ actions:
   {deploy,subscribe,unsubscribe,discover-log-groups}
                         Actions to be executed by the script:
     deploy              Deploy AWS infrastructure (AWS Kinesis Firehose and AWS Lambda) configured to forward logs from AWS CloudWatch to Dynatrace ActiveGate.
+    delete              Delete AWS infrastructure.
```
mmajcher, line 37: " Delete AWS infrastructure." sounds a little scary

.travis.yml
```diff
@@ -1,31 +1,65 @@
...
         - docker run --rm lambda_image bash -c 'python3 -m pylint --rcfile=pipeline/pylint.cfg $(find src/ -name "*.py")'
         - docker run --rm lambda_image bash -c 'cd src && python3 -m pytest -v ../tests/unit'
+    - stage: build
+      name: Build AWS Lambda package
+      #if: tag =~ /^release.*$/
+      language: shell
+      services: docker
+      script:
+        - docker run --rm -v `pwd`:/var/task amazon/aws-sam-cli-build-image-python3.8 bash -c './build-release-package.sh'
+        - cp -p dynatrace-aws-log-forwarder.zip ~/build-stages-shared-storage/$TRAVIS_BUILD_NUMBER
```
mmajcher, line 37: How does this work? When is stuff from ~/build-stages-shared-storage pushed to s3?

mmajcher: Ok, please confirm if I understand correctly. "before_install" above which installs aws-cli and does s3-sync is executed before each step?
But still then, when is this stuff here pushed to s3? (we need some push after this step here, right?)
Also, could before_install use some docker image so we don't run install aws-cli from zero each time?

.travis.yml
mmajcher: Hah, I got a typo in word "typo" :) :)

dynatrace-aws-logs.sh
kamiljsokolowski: We already discussed this in one of our daily meetings.

IMHO the customer needs to have an option to delete the log forwarder. Update the docs YES, but add a note about the need to remove all log group subscriptions for the given stack. Otherwise, we are duplicating code.

dynatrace-aws-logs.sh
kamiljsokolowski: Ahh.

dynatrace-aws-logs.sh
mmajcher: I think the "delete" option should do this log group subscriptions on its own. 
1) Autodiscover all 2) Pass this to unsubscribe.
This can be easily automated I think and should be from UX perspective

.travis.yml
kamiljsokolowski: I'm not using the script from the repo. All the archives/packages used are products of the current build.

1. I need to copy the entire release package from the build step to the shared storage so that it is available later for the deploy stage.
2. For e2e I just need the Lambda package from the build step, as both the script and the CF template are parts of the repo. To zip all parts, copy them and unzip them in the next step would be overengineering and cause trouble (overwriting repo files)

.travis.yml
kamiljsokolowski: Please see my [comment](https://github.com/dynatrace-oss/dynatrace-aws-log-forwarder/pull/18#discussion_r660463977) above;)

.travis.yml
kamiljsokolowski: Ahh again.

.travis.yml
mmajcher: Let's say someone does some change to build-release-package.sh and in result one of files is not included in the release zip. The test would not find this.
E2E test should only use release resources, not touch anything from repo to. To avoid overwrite you can do this unzip in new temp folder. 
E2E idea is not to deploy the thing as easily as possible but to deploy the thing using 100% same resources as client is using, in this case only release.zip

mmajcher: `To zip all parts, copy them and unzip them in the next step would be overengineering and cause trouble (overwriting repo files)` you don't have to zip them and copy, you already have the artifact from build step

dynatrace-aws-logs.sh
kamiljsokolowski: Maybe but IMHO it's consistent with `deploy` description.

mmajcher: My proposition:
```
deploy - Deploy AWS log forwarder stack to AWS.
delete - Delete AWS log forwared stack.
```
Actually deploy description is not really even up to date (actually it's my fault ;) ) as there are more elements to this infrastructure now (ec2, vpc, etc. )

.travis.yml
kamiljsokolowski: Yes, `before_install` and `after_success` are executed with each stage. This is mainly because those steps do not take too much time and applying them to each step (`build`, `e2e_tests` and `deployment`) independently would produce repeatable code. The S3 sync is exected before each stage for the packages to be downloaded and after for the build products to be pushed.

As for the Docker image, AFAIK Travis does not allow for custom images to be used as build agents.

dynatrace-aws-logs.sh
kamiljsokolowski: I would consider two things here:
1. Is there a scenario, where the end-user does want to delete the log forwarder but keep the logs and subscription groups? If not, then we could add the unsubscribe function.
2. Adding `unsubscribe` as part of `delete` would IMHO make the latter inconsistent with 'deploy' and could confuse the user.

dynatrace-aws-logs.sh
kamiljsokolowski: Sounds good. Updating the description then.

dynatrace-aws-logs.sh
mmajcher: 1. No, we don't have such scenario 
2. "delete" is almost useless for the user from my point of view. We can name it "purge" or "uninstall" or something else, but I expect full cleanup from this step.
My concern is following: I am a user, you gave me "delete" option, I expect it to delete properly and not leave lots of mess behind. If you want to give me "delete" option which only does like 30% of work then it's better not to give this option

.travis.yml
mmajcher, line 47: Is $TRAVIS_BUILD_NUMBER unique globally?

.travis.yml
mmajcher: Ok, I see it now! Very nice. I was confused by `after_success` since it is thrown out at the bottom and somehow I intuitively thought it is executed only once after successful build.

.travis.yml
kamiljsokolowski: Fixed.

dynatrace-aws-logs.sh
kamiljsokolowski: Fixed.

dynatrace-aws-logs.sh
kamiljsokolowski: Done.

.travis.yml
```diff
@@ -1,31 +1,65 @@
...
+      if: branch = master OR type = pull_request
+      language: shell
+      services: docker
+      before_script: cd ~/build-stages-shared-storage/$TRAVIS_BUILD_NUMBER
+      script:
+        - unzip dynatrace-aws-log-forwarder.zip
+        # deploy log forwarder resources reqiured for e2e to AWS
+        - ./dynatrace-aws-logs.sh deploy --target-url $DYNATRACE_ENV_URL --target-api-token $DYNATRACE_API_KEY --use-existing-active-gate true --require-valid-certificate $VERIFY_SSL --stack-name aws-log-forwarder-e2e-build-$TRAVIS_BUILD_NUMBER
+        # delete log forwarder AWS resources
+        - aws cloudformation delete-stack --stack-name aws-log-forwarder-e2e-build-$TRAVIS_BUILD_NUMBER
```
Dawid-Kaszubski, line 49: [Minor] Deleting could be in `after_script` part

Dawid-Kaszubski (approved):

.travis.yml
kamiljsokolowski: Updated the pipeline, so that e2e only utilizes build package content.

dynatrace-aws-logs.sh
kamiljsokolowski: N/A

.travis.yml
kamiljsokolowski: N/A

.travis.yml
kamiljsokolowski: Yes if we are thinking in terms of repo scope.

dynatrace-aws-logs.sh
kamiljsokolowski: `delete` function has been removed and the execution replaced by AWS CLI call.

.travis.yml
kamiljsokolowski: Good catch, but I should probably use `after_success` so that, when something goes wrong, the created resources will be available for debugging.

mmajcher: I like the idea but on the other hand you have the logs of the stack you created and then deleted. (aws log forwader logs)
If we leave the resources then they might not get cleaned up. I am not strong on either way

.travis.yml
mmajcher: Can you give an example of its value? 
I would add a timestamp to the name, just as we do in integration test for AWS Metric Streaming. We did this to ensure unique names. Seems not really needed here but could add this just to be sure. But it's up to you

.travis.yml
```diff
@@ -1,31 +1,65 @@
...
+      services: docker
+      script:
+        - docker run --rm -v `pwd`:/var/task amazon/aws-sam-cli-build-image-python3.8 bash -c './build-release-package.sh'
+        - cp -p dynatrace-aws-log-forwarder.zip ~/build-stages-shared-storage/$TRAVIS_BUILD_NUMBER
+    - stage: e2e_tests
+      name: Execute e2e tests
+      if: branch = master OR type = pull_request
+      language: shell
+      services: docker
+      before_script: cd ~/build-stages-shared-storage/$TRAVIS_BUILD_NUMBER
```
mmajcher, line 43: If you go to shared storage. And then unzip things in there. And then "after_success" does s3 sync. Possibly you will send things from zip to s3?
I would go for new special sterile directory ;)

mmajcher (dismissed):

.travis.yml
kamiljsokolowski: The value is simply the build number i.e. 171. It was the easiest unique number to base ID on because it is stored as env variable by Travis.
If we have a common best practice in generating unique IDs, I'm more than happy to comply;)

mmajcher: Nah, we don't have any common practices. Seems good. 
In previous tests we were using jenkins and there build number was not unique, each branch has its own numbering and conflicts could have come up

.travis.yml
kamiljsokolowski: Let us set this aside for when the test part is ready. For now, if as you say CF logs get the debugging job done, `after_script` will do.

.travis.yml
kamiljsokolowski: Moved away from shared storage to a dedicated directory.

mmajcher (approved):

PR #20, opened 2021-07-09
APM-306464

kamiljsokolowski:
This update will add e2e test execution.


.travis.yml
```diff
@@ -32,38 +32,70 @@ jobs:
...
+        - mkdir -p $TRAVIS_BUILD_SHARED_STORAGE
+        - aws s3 sync $AWS_S3_BUILD_SHARED_STORAGE $TRAVIS_BUILD_SHARED_STORAGE
+        - cp -p $TRAVIS_BUILD_SHARED_STORAGE/dynatrace-aws-log-forwarder.zip ./tests/e2e
+        - cd ./tests/e2e
         - unzip dynatrace-aws-log-forwarder.zip
       script:
         # deploy log forwarder resources reqiured for e2e to AWS
-        - ./dynatrace-aws-logs.sh deploy --target-url $DYNATRACE_ENV_URL --target-api-token $DYNATRACE_API_KEY --use-existing-active-gate true --require-valid-certificate $VERIFY_SSL --stack-name aws-log-forwarder-e2e-build-$TRAVIS_BUILD_NUMBER
+        - ./dynatrace-aws-logs.sh deploy --target-url $DYNATRACE_ENV_URL --target-api-token $DYNATRACE_API_KEY --use-existing-active-gate true --require-valid-certificate $VERIFY_SSL --stack-name $AWS_CLOUDFORMATION_STACK
+        # create AWS CloudWatch Logs log groups and assign log streams
```
mmajcher, line 66: Those descriptive comments are not needed, very easy to read this from the code

tests/e2e/requirements.txt
```diff
@@ -0,0 +1,5 @@
+###### Requirements without Version Specifiers ######
```
mmajcher, line 1: Please put specific versions

.travis.yml
dt-be: For me, let them be, we suffer from too few comments in that repo as it is.

mmajcher: Comments of type "create log groups" where next line says "aws logs create-log-group" bring no value at all from my perspective. I am all for comments but their role is to bring context and explain things, not repeat what next line says

dt-be: Brevity of comment vs the actual mouthful to read in the next lines makes a difference to me. Just making a second point here, don't intend to discuss it much further.

.travis.yml
```diff
@@ -32,38 +32,70 @@ jobs:
...
         - docker run --rm -v `pwd`:/var/task amazon/aws-sam-cli-build-image-python3.8 bash -c './build-release-package.sh'
-        - cp -p dynatrace-aws-log-forwarder.zip ~/build-stages-shared-storage/$TRAVIS_BUILD_NUMBER
+      after_success:
+        - cp -p dynatrace-aws-log-forwarder.zip $TRAVIS_BUILD_SHARED_STORAGE
+        - aws s3 sync $TRAVIS_BUILD_SHARED_STORAGE $AWS_S3_BUILD_SHARED_STORAGE
     - stage: e2e_tests
       name: Execute e2e tests
       if: branch = master OR type = pull_request
       language: shell
       services: docker
```
wojtasskorcz, line 50: docker is not needed here

.travis.yml
```diff
@@ -32,38 +32,70 @@ jobs:
...
         # delete log forwarder AWS resources
-        - aws cloudformation delete-stack --stack-name aws-log-forwarder-e2e-build-$TRAVIS_BUILD_NUMBER
+        - AWS_CLOUDFORMATION_STACK_DELIVERY_S3_BUCKET=$(aws cloudformation list-stack-resources --stack $AWS_CLOUDFORMATION_STACK --query "StackResourceSummaries[?contains(LogicalResourceId, 'DeliveryBucket')].PhysicalResourceId" --output text)
+        - aws s3 rm s3://${AWS_CLOUDFORMATION_STACK_DELIVERY_S3_BUCKET} --recursive
+        - aws cloudformation delete-stack --stack-name $AWS_CLOUDFORMATION_STACK
     - stage: deployment
       name: Github Release Deployment
       if: tag =~ /^release.*$/
       language: shell
       services: docker
```
wojtasskorcz, line 86: Not needed here either. Please update to the latest master -- I removed it there already

.travis.yml
wojtasskorcz: If at least two people find them useful (Beata and Kamil), I'd leave them. It's not like they obscure the code or anything. They may be not as useful as
```
# include $HOME/.local/bin for `aws`
```
but I think they do no harm

tests/e2e/main.py
```diff
@@ -0,0 +1,124 @@
+#!/usr/bin/env python3
```
wojtasskorcz, line 1: You could also add the license info here

tests/e2e/main.py
```diff
@@ -0,0 +1,124 @@
...
+                    'timestamp': epoch_timestamp_in_ms,
+                    'message': message_content
+                },
+            ],
+        )
+        print("Generated log event: %s" % log_event)
+    except ClientError as e:
+        if e.response['Error']['Code'] == 'DataAlreadyAcceptedException' or e.response['Error']['Code'] == 'InvalidSequenceTokenException':
+            print("Given log batch has already been accepeted. Retrying with the following sequence token: " + e.response['expectedSequenceToken'])
+            log_event_retry = aws_client.put_log_events(
```
karolkieglerski, line 33: this is semi e2e test because you own generate log, better solution will be to generate log by some service like Lambda ora ECS or something else

tests/e2e/main.py
```diff
@@ -0,0 +1,124 @@
...
+    dynatrace_log_records_search_results = search_dynatrace_log_records(args.url_prefix,message_content)['results']
+    time_elapsed = 0
+    while bool(dynatrace_log_records_search_results) is False:
+        print('Waiting for log events to be picked up by the cluster...')
+        time.sleep(10)
+        time_elapsed = time_elapsed + 10
+        dynatrace_log_records_search_results = search_dynatrace_log_records(args.url_prefix,message_content)['results']
+        # Fail the test after 5min
+        if time_elapsed == 300:
+            print('Generated logs where not found. End-to-end test has failed.')
```
wojtasskorcz, line 119: typo in "where"

tests/e2e/main.py
```diff
@@ -0,0 +1,124 @@
...
+    # Generate a test log event
+    generate_log_event(args.log_group_name,
+                       args.log_stream_name)
+
+    # Search Dynatrace logs for records with specific message content and timestamp
+    dynatrace_log_records_search_results = search_dynatrace_log_records(args.url_prefix,message_content)['results']
+    time_elapsed = 0
+    while bool(dynatrace_log_records_search_results) is False:
+        print('Waiting for log events to be picked up by the cluster...')
+        time.sleep(10)
```
wojtasskorcz, line 114: It's a bit different approach than described in the story:
```
use retry behavior to start verification early but if delay is bigger then expected then retry few times with e.g. 1 minute wait; e.g. initial wait 5 minutes and then 5 retries after 1 minute until test succeeds
```
Are we okay with that?

tests/e2e/main.py
```diff
@@ -0,0 +1,124 @@
...
+    request_headers       = {'Authorization': 'Api-Token ' + args.api_token,
+                            'Content-Type': 'application/json',
+                            'charset': 'utf-8'}
+    message_content       = 'Dynatrace-AWS-Log-Forwarder-End-to-End-test-' + args.unique_message_content
+
+    # Generate a test log event
+    generate_log_event(args.log_group_name,
+                       args.log_stream_name)
+
+    # Search Dynatrace logs for records with specific message content and timestamp
```
karolkieglerski, line 109: instead of writing your own method to test use pytest as we have in 2 other projects. pytest is used also for unit and integration

tests/e2e/main.py
```diff
@@ -0,0 +1,124 @@
...
+            logEvents=[
+                {
+                    'timestamp': epoch_timestamp_in_ms,
+                    'message': message_content
+                },
+            ],
+        )
+        print("Generated log event: %s" % log_event)
+    except ClientError as e:
+        if e.response['Error']['Code'] == 'DataAlreadyAcceptedException' or e.response['Error']['Code'] == 'InvalidSequenceTokenException':
```
wojtasskorcz, line 31: Is it an expected situation? The documentation says:
```
An upload in a newly created log stream does not require a sequence token
```
https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_PutLogEvents.html

and here we have a new log stream, right?

wojtasskorcz: Ahh, just realized -- this test can be run outisde of Travis on an existing log stream. Is this the reason to include this logic?

tests/e2e/main.py
mmajcher: We discussed this and we are perfectly fine with this solution. External service brings very little difference for what we want to test so this is mostly theoretical thing. Our solution starts at cloudwatch log group and ends in cluster API.

tests/e2e/main.py
mmajcher: I think it is fine, do you see any reason against?

wojtasskorcz: Nah, I'm good. Just wanted to attract your attention to it because maybe someone else would have a strong opinion. Let's keep it as is then

tests/e2e/main.py
kamiljsokolowski: As Marek mentioned, MVP was to generate a log without a specific source just to test the forwarding solution.

tests/e2e/main.py
kamiljsokolowski: 🥴

tests/e2e/main.py
kamiljsokolowski: I've tested this in several iterations and, if the forwarder was deployed properly, the delay was never bigger than 2min. However, this value could be subject to change when we start using the dedicated AG solution.

tests/e2e/main.py
kamiljsokolowski: Exactly:)

tests/e2e/main.py
kamiljsokolowski: Done.

tests/e2e/main.py
kamiljsokolowski: Done.

tests/e2e/requirements.txt
kamiljsokolowski: Done.

.travis.yml
```diff
@@ -32,38 +32,70 @@ jobs:
...
+        # set up shared storage for build artifacts
+        - mkdir -p $TRAVIS_BUILD_SHARED_STORAGE
+        - aws s3 sync $AWS_S3_BUILD_SHARED_STORAGE $TRAVIS_BUILD_SHARED_STORAGE
+        - cp -p $TRAVIS_BUILD_SHARED_STORAGE/dynatrace-aws-log-forwarder.zip ./tests/e2e
+        - cd ./tests/e2e
         - unzip dynatrace-aws-log-forwarder.zip
       script:
         # deploy log forwarder resources reqiured for e2e to AWS
-        - ./dynatrace-aws-logs.sh deploy --target-url $DYNATRACE_ENV_URL --target-api-token $DYNATRACE_API_KEY --use-existing-active-gate true --require-valid-certificate $VERIFY_SSL --stack-name aws-log-forwarder-e2e-build-$TRAVIS_BUILD_NUMBER
+        - ./dynatrace-aws-logs.sh deploy --target-url $DYNATRACE_ENV_URL --target-api-token $DYNATRACE_API_KEY --use-existing-active-gate true --require-valid-certificate $VERIFY_SSL --stack-name $AWS_CLOUDFORMATION_STACK
```
dt-be, line 65: I personally like exporting myself those repeating dynatrace-aws-logs variables 
```
export STACK_NAME=BeluPerfLogsCLientSecond
# more exports
./dynatrace-aws-logs.sh deploy --use-existing-active-gate true

export LOG_GROUPS_LIST="/aws/lambda/BeluAWSLogsPerformanceTest_inputLogsProducingLambda"
./dynatrace-aws-logs.sh subscribe
```

I thought maybe you could set them once here or even call the global like that already and it could work. 

We get more brevity, maybe a little bit more readability because of it, but for the price of it maybe not being explicit from this file here where the variable is used. Just a suggestion, but this is totally optional.

.travis.yml
```diff
@@ -1,7 +1,13 @@
+dist: focal
 env:
   global:
     # include $HOME/.local/bin for `aws`
     - PATH=$HOME/.local/bin:$PATH
+    - TRAVIS_BUILD_SHARED_STORAGE="&{TRAVIS_BUILD_DIR}/build-stages-shared-storage/${TRAVIS_BUILD_NUMBER}"
+    - AWS_S3_BUILD_SHARED_STORAGE="s3://aws-log-forwarder-travis-build-shared-storage/${TRAVIS_BUILD_NUMBER}"
```
dt-be, line 7: I really appreciate the paths landing in vars rather than strings all over the place, makes it so much easier to read 👍

.travis.yml
mmajcher: Command line arguments is the main deployment way from the docs and for me this would be a reason to stick to that.
If for some reason deployment with exports does not work then its not as big as a problem as if main deployment way with cmd args stops working

dt-be: Ok, forgot there's anything in between that could go wrong.

.travis.yml
kamiljsokolowski: Removed.

.travis.yml
kamiljsokolowski: Done.

tests/e2e/main.py
```diff
@@ -0,0 +1,138 @@
...
+            print("Given log batch has already been accepeted. Retrying with the following sequence token: " + e.response['expectedSequenceToken'])
```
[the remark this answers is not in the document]
kamiljsokolowski: Done.

tests/e2e/main.py
kamiljsokolowski: We had a discussion concerning the above during today's daily and decided to stick with the current approach. In our case re-writing to `pytest`, considering a more complicated script and my lack of knowledge on the subject of using the lib for scenarios other, than unit testing, would require more effort.

We could create a task for a future re-factor/improvement.

.travis.yml
```diff
@@ -32,38 +32,70 @@ jobs:
       #if: tag =~ /^release.*$/
       language: shell
       services: docker
+      install:
+        # set up awscli packages
+        - python3 -m pip install --user awscli
```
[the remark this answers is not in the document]
kamiljsokolowski: This does install AWS CLI as Python module. Frankly, this is the easiest way to do it (as per Travis docs). I've added a specific v1 version (`1.20.9`) to be installed.

Side note: one can't install AWS CLI v2 via pip because it is not available in PyPi (some arbitrary AWS decision).

.travis.yml
```diff
@@ -32,38 +32,71 @@ jobs:
...
         - docker run --rm -v `pwd`:/var/task amazon/aws-sam-cli-build-image-python3.8 bash -c './build-release-package.sh'
-        - cp -p dynatrace-aws-log-forwarder.zip ~/build-stages-shared-storage/$TRAVIS_BUILD_NUMBER
+      after_success:
+        - cp -p dynatrace-aws-log-forwarder.zip $TRAVIS_BUILD_SHARED_STORAGE
+        - aws s3 sync $TRAVIS_BUILD_SHARED_STORAGE $AWS_S3_BUILD_SHARED_STORAGE
     - stage: e2e_tests
       name: Execute e2e tests
-      if: branch = master OR type = pull_request
+      # if: branch = master OR type = pull_request
+      if: branch = APM-306464
```
wojtasskorcz, line 49: hmm?

.travis.yml
wojtasskorcz: Doesn't seem like it. I can still see it in the diff

.travis.yml
wojtasskorcz: Same as above, I can still see it in the diff

wojtasskorcz (changes requested):
See new comments in the resolved conversations regarding `services: docker`

.travis.yml
kamiljsokolowski: Please see the commit message. This one is for debugging purposes only (normally e2e would run only on `master` or for a PR). I'll switch it back as soon as the build passes.

wojtasskorcz (dismissed):

dt-be (approved):

PR #24, opened 2021-07-14
APM-313572 validate ActiveGate state before starting deployment

sq2gxo:


dynatrace-aws-logs.sh
```diff
@@ -91,6 +91,19 @@ arguments:
...
+  function check_activegate_state() {
+    if ACTIVE_GATE_STATE=$(curl -ksS "${TARGET_URL}/rest/health" --connect-timeout 20); then
+      if [[ "$ACTIVE_GATE_STATE" != "RUNNING" ]]
+      then
+        echo -e ""
+        echo -e "\e[91mERROR: \e[37mActiveGate endpoint is not reporting RUNNING state. Please verify provided values for parameter: --target-url (${TARGET_URL})."
+        exit 1
+      fi
+    else
+        echo -e "\e[93mWARNING: \e[37mFailed to connect to ActiveGate url $TARGET_URL to check state. It can be ignored if ActiveGate does not allow public access."
```
mmajcher, line 103: For AWS, existingAG option requires AG with public access. So this would be error + exit 1

dynatrace-aws-logs.sh
```diff
@@ -214,11 +227,17 @@ EOF
   fi
 
   TARGET_URL=$(echo "$TARGET_URL" | sed 's:/*$::')
+
+  if [[ "$USE_EXISTING_ACTIVE_GATE" == "true" ]]; then
+    check_activegate_state
+  fi
```
dt-be, line 233: I'd move it 239 after Marysia's check_log_ingest_url - it seems to me to make sense to verify the TARGET_URL format first  & there's the same if

dynatrace-aws-logs.sh
sq2gxo: why then in line 120 we do not have exit?
should I change both places?

mmajcher: Good question, I think in 120 the thought was that AG could be only exposing publicly one endpoint (for logs ingest). So following this approach this here should be a Warning.
To be honest I am not sure what approach we should take. It is tempting but also not most secure to ask for full public access. 
What do you think?

I think @mswiatkowska added somewhere another check that sends test log to given existingAG, is this check then not a duplicate?
Anyhow, assuming we have this check I described here it seems safe to leave this Warning as is.

dynatrace-aws-logs.sh
sq2gxo: Verification of URL pattern is in line 224 - above.

I think current order is correct as it starts from basic check into final log ingest check (next check is depending on conditions checked in previous one):
- check ActiveGte status
- check token permission
- try to ingest test log line

dynatrace-aws-logs.sh
sq2gxo: there are three check in case of using existing activegate to provide more detailed feedback to user what is not ok,
starting from basic check into final log ingest check (next check result depends on conditions checked in previous one):
check ActiveGte status (check_activegate_state)
check token permission (check_api_token)
try to ingest test log line (check_log_ingest_url)

mmajcher: OK this looks very nice. We have error+exit 1 on check_log_ingest_url and it is 100% enough. So this line here is good as warning 👍

mmajcher:
Just FYI, those improvements seem really needed: 
https://dev-jira.dynatrace.org/browse/APM-313681
Customer set up AG but did not assign public IP and did not expose to the internet

dynatrace-aws-logs.sh
dt-be: Missed it, thanks.

dt-be (approved):

mmajcher (approved):

mswiatkowska (approved):

wojtasskorcz (approved):

PR #26, opened 2021-07-15
APM-313830: Disable SFM metrics push. It does not work with newAG opt…

mmajcher:
…ion at all! Lambda in VPC cannot connect to Cloudwatch at the moment.

SFM push will work as usual in existing AG setup.

It does not work in newAG setup. SFM putMetricData times out after 5 minutes (in a setup where we tweaked lambda timeout to 10 minutes; normal lambda timeout in our stack is 1 minute so lambda will time out before putMetricData times out)
Our VPC design is following: EC2 has public IP (can connect everything) and is protected by security group, only allows traffic from Lambda. 
Lambda is connect to VPC but gets no IP.
Since the subnet is public subnet and does not have NAT gateway, lambda can not obtain any IP when it reaches out to internet. So the request to public Cloudwatch endpoint is going out but response never arrives back.

So actually logs should be streaming fine (happens before putMetricData is called) but lambda will take lots of time (5 minutes putMetricData SDK timeout).

For now, I disabled SFM putMetricData at all. 
In coming days we will discuss if and how we want to redesign VPC setup to allow lambda putMetricData push. 
The solution would be to put lambda in additional private subnet that will go to internet through NAT gateway in public subnet (this way an entity without IP can send request and receive response - using NAT gw as intermediary).

Please approve, merge & release ASAP

### DIFF
--- a/src/index.py
+++ b/src/index.py
@@ -55,10 +55,10 @@ def handler(event, lambda_context):
         log_error_with_stacktrace(e, "Exception caught in top-level handler")
         result = TransformationResult.ProcessingFailed
 
-    try:
-        context.sfm.push_sfm_to_cloudwatch()
-    except Exception as e:
-        log_error_with_stacktrace(e, "SelfMonitoring push to Cloudwatch failed")
+    # try:
+    #     context.sfm.push_sfm_to_cloudwatch()
+    # except Exception as e:
+    #     log_error_with_stacktrace(e, "SelfMonitoring push to Cloudwatch failed")
 
     return kinesis_data_transformation_response(records, result)
 


wojtasskorcz (approved):

src/index.py
```diff
@@ -55,10 +55,10 @@ def handler(event, lambda_context):
...
         result = TransformationResult.ProcessingFailed
 
-    try:
-        context.sfm.push_sfm_to_cloudwatch()
-    except Exception as e:
-        log_error_with_stacktrace(e, "SelfMonitoring push to Cloudwatch failed")
+    # try:
+    #     context.sfm.push_sfm_to_cloudwatch()
+    # except Exception as e:
+    #     log_error_with_stacktrace(e, "SelfMonitoring push to Cloudwatch failed")
```
dt-be, line 61: You could have commented it out inside of sfm and we would have them logged at least. But will do for today.

```
    def push_sfm_to_cloudwatch(self):
        metrics = self._generate_metrics()
        # cloudwatch = boto3.client('cloudwatch')
        # try:
        #     for i in range(0, len(metrics), 20):
        #         metrics_batch = metrics[i:(i + 20)]
        #         cloudwatch.put_metric_data(MetricData=metrics_batch, Namespace='DT/LogsStreaming')
        # except Exception as e:
        print("Print metrics on SFM push failure: " + str(metrics))
            # raise e
```

dt-be (approved):

src/index.py
mmajcher: They are still not good to be logged I think. Because right now SFM batch is getting very large when lots of records are coming in one batch (you can see example in this ticket's comments). We will also fix this in a simple way but it's not there yet

mmajcher: And they are only logged on sfm push failure I think (not sure right now)

dt-be: >And they are only logged on sfm push failure I think (not sure right now)
Yes, I've pasted an edited version above that would log it always.
> Because right now SFM batch is getting very large
Yes, I remember, that's why won't discuss it today. Just commented to have a snippet ready for tomorrow's discussions.

PR #28, opened 2021-07-22
APM-312886

equardo:
Small change in the CloudFormation template making it so the name of EC2 containing ActiveGate is named *stack-name*-active-gate.

### DIFF
--- a/dynatrace-aws-log-forwarder-template.yaml
+++ b/dynatrace-aws-log-forwarder-template.yaml
@@ -578,6 +578,9 @@ Resources:
             DeleteOnTermination: "true"
             VolumeSize: "8"
       SecurityGroupIds: [ !Ref VPCSecurityGroup ]
+      Tags:
+        - Key: Name
+          Value: !Join [ "-", [ !Ref 'AWS::StackName', "active-gate" ] ]
       UserData:
         Fn::Base64:
           !Sub


wojtasskorcz (approved):

mmajcher (approved):
Very nice! 
Glad we managed to secure all accesses :)

dynatrace-aws-log-forwarder-template.yaml
```diff
@@ -578,6 +578,9 @@ Resources:
             DeleteOnTermination: "true"
             VolumeSize: "8"
       SecurityGroupIds: [ !Ref VPCSecurityGroup ]
+      Tags:
+        - Key: Name
+          Value: !Join [ "-", [ !Ref 'AWS::StackName', "active-gate" ] ]
```
dt-be, line 583: Funny that name's a tag. I assume it works:P

dt-be (approved):

dynatrace-aws-log-forwarder-template.yaml
equardo: image

equardo: It is kind of weird how it's a tag but from what I can gather that's how it is and it does work.

mmajcher: Can confirm, it is a tag

PR #31, opened 2021-07-26
APM-315548 adjust DYNATRACE_URL_REGEX to accept also managed enviroents

sq2gxo:


dynatrace-aws-logs.sh
```diff
@@ -18,7 +18,7 @@ readonly LAMBDA_ZIP_NAME="dynatrace-aws-log-forwarder-lambda.zip"
 
 readonly DEFAULT_STACK_NAME="dynatrace-aws-logs"
 
-readonly DYNATRACE_TARGET_URL_REGEX="^https:\/\/[-a-zA-Z0-9@:%._+~=]{1,256}[\/]{0,1}$"
+readonly DYNATRACE_TARGET_URL_REGEX="^(https?:\/\/[-a-zA-Z0-9@:%._+~=]{1,256}\/?)(\/e\/[a-z0-9-]{36}\/?)?$"
```
mmajcher, line 21: They have 36 chars for ID after "e" part?

mmajcher (approved):

dynatrace-aws-logs.sh
sq2gxo: Yes, Dynatrace environment ID for managed installations is in UUID format

mswiatkowska (approved):

dynatrace-aws-logs.sh
dt-be: Is only Managed there, though, i.e. always 36 chars? AG check uses {1,36} not exactly {36}

This is what I have for my jxw setup and it doesn't match, but maybe I miss something elsewhere:

export TARGET_URL=https://ec2-11-222-333-44.compute-1.amazonaws.com:9999/e/jxw01498
image
image

sq2gxo: Dynatrace environment ID convention for Environment ID is:
- SaaS: 8 chars (three letters, five digits)
- Managed 36 chars

for ACTIVE_GATE_TARGET_URL_REGEX weare accepting from 1 to 36 characters, but actually it will also be ok to accept 8 or 36 only

https://ec2-11-222-333-44.compute-1.amazonaws.com:9999/e/jxw01498 is an ActiveGate URL - why you want to match it to Dynatrace environment URL?


- SaaS -

dt-be: I had some trouble seeing both regexes on github and wanted to ask and make sure either way. I only later pulled the changes locally and could see there is a separate regex for cluster and AG. So what we're matching against is e.g.:

`https://11111111-3333-4444-7777-333333333333-1.managed.internal.dynatrace.com:8021/e/55555555-6666-4444-9999-dddddddddddd/` for Managed
vs just
`https://klm56789.live.dynatrace.com/` for Saas

dt-be (approved):

PR #29, opened 2021-07-22
APM-312140 - Added parser for AWS App Runner logs

bmrozinski-dt:
- application logs and service logs are covered
- also added attributes checks to unit test

PR #32, opened 2021-07-26
APM-313844: Log rules for mysql

mswiatkowska:
Logs examples can be found on jxw


src/config/rds_mysql.json
```diff
@@ -0,0 +1,44 @@
...
+          "key": "aws.service",
+          "pattern": "'rds'"
+        },
+        {
+          "key": "aws.resource.id",
+          "pattern": "resource_id"
+        },
+        {
+          "key": "aws.arn",
+          "pattern": "format_arn('arn:{}:rds:{}:{}:db:{}', [partition, region, account_id, resource_id])"
```
dt-be, line 27: I missed that in Bartek's PR in the end - could you add `priority` to this rule and set it to 1 or 10, and then reuse the aws.arn in dt.source_entity calculation? It probably has minor impact, but still spares us one re-defintion and one re-calculation of the arn. Something like this should work:

```
{
          "key": "aws.arn",
          "pattern": "format_arn('arn:{}:rds:{}:{}:db:{}', [partition, region, account_id, resource_id])",
          "priority": 10
        },
        {
          "key": "dt.source_entity",
          "pattern": "dt_meid_rds_v2(aws.arn))"
        },       
```

mmajcher: I think in Bartek's PR there was a slight difference, i.e meid input was little different from ARN. Actually we have been looking at this in the office :)

dt-be: Ok, great - I just remembered I wanted to check if this was same or two different, but never did in the end. So just here then, here it's two same ones.

tests/unit/logs/test_transformation_full.py
```diff
@@ -503,6 +503,72 @@
...
+            "logGroup": "/aws/rds/instance/mysql-db-for-logs/audit",
+            "logStream": "mysql-db-for-logs",
+            "messageType": "DATA_MESSAGE",
+            "owner": "444000444",
+            "subscriptionFilters": ["mysql-audit-filter"],
+            "logEvents": [
+                {
+                    "id": "35958590510527767165636549608812769529777864588249006080",
+                    "timestamp": "12345",
+                    "message": "20210722 17:15:28,ip-10-1-3-230,admin,157.25.19.100,13,0,DISCONNECT,testdb,,0,SSL"
```
dt-be, line 517: Just to make sure as we looked at this code earlier - this `testdb` here (that sometimes is present, sometimes isn't) is not the just some internal mysql database name that we don't need. And the actual AWS instance name in which we're interested in to have a link to entity is already there in logGroup.

dt-be:
Any chance you could add the examples files like Bartek did?
https://github.com/dynatrace-oss/dynatrace-aws-log-forwarder/pull/29/files#diff-c664b9130fcf8f1826b2e47809d104747458e43974ec36f5d25ee607f86a7d9c

I know there is a test, but test is just a singe entry of each logType, would be nice to have at least a couple of and I think I saw you already had something like it.

src/config/rds_mysql.json
mswiatkowska: Done

dt-be (approved):

sq2gxo (dismissed):

sq2gxo:
Rebased branch to fix merge conflicts, please re-review

sq2gxo (approved):

tests/unit/logs/test_transformation_full.py
wojtasskorcz: Couldn't understand this comment at first. This one superfluous `not` is super confusing, especially that the rest of the sentence has entirely different meaning then :)

examples/rds.txt
```diff
@@ -0,0 +1,111 @@
...
+            },
+            {
+                "id": "36282733332258008410023818770931443951379879892065320960",
+                "timestamp": 1626974032000,
+                "message": "20210726 12:05:10,ip-10-1-3-230,db_user,157.25.19.100,1029,242734,QUERY,`testdb`,'select * from mysql.user\nLIMIT 0, 1000',1142,,"
+            },
+            {
+                "id": "36282735473129547468963640358360398125622584466185256962",
+                "timestamp": 1626974128000,
+                "message": "20210722 17:15:28,ip-10-1-3-230,admin,157.25.19.100,13,0,DISCONNECT,testdb,,0,SSL"
```
wojtasskorcz, line 24: AWS can't even be consistent when it comes to having dbname in backticks or not. Love this cloud

src/config/rds_mysql.json
```diff
@@ -0,0 +1,45 @@
...
+          "pattern": "format_arn('arn:{}:rds:{}:{}:db:{}', [partition, region, account_id, resource_id])",
+          "priority": 10
+        },
+        {
+          "key": "dt.source_entity",
+          "pattern": "dt_meid_rds_v2(\"aws.arn\")"
+        },
+        {
+          "key": "severity",
+          "pattern": "if(log_type == 'slowquery' || contains(log_content, '[Warning]'), &'WARNING', &if(contains(log_content, '[Error]'), &'ERROR', &'INFO', @), @)"
```
wojtasskorcz, line 36: From `examples/rds.txt` it looks like slowquery logs have `logGroup` ending in `/general` but in `tests/unit/logs/test_transformation_full.py` the `logGroup` name ends with `/slowquery`. Which one is correct?

wojtasskorcz: Also, I remember one discussion in our team that we'd like to avoid using `contains` due to performance reasons and only use `starts_with`. @bmrozinski-dt I think it was related to one of your PRs. Do you remember the details?

bmrozinski-dt: We decided to take a safe path and we determine severity only based on the `starts_with` condition and avoid the `contains` because of performance (single log content can contain even thousands of characters). We haven't done any performance tests though. So maybe `contains` is not such a risk...
If there is a way to check only the first 50 characters (for example) - maybe it could be a more efficient way.

dt-be: @mswiatkowska already also had this discussion with @mmajcher I believe, and did profiling of contains for those particular log type as a result. I expected she came back with the results, but it would make sense to have them publicly too and available to all - hopefully she can confirm when she comes back. 

What she told me was, for those logs characteristics:
* with fixed schema (as in the examples) that also limits expected average length
* (my assumption) a more consistent test sample (possibly unlike the profiling I already did in CloudTrail before, that was deemed insufficient and failed to persuade you)
* 10k logs > contains is more expensive by 0,6s (3.2s vs 2.6s)
* 1k logs > contains is more expensive by 0.1s

For me this difference is worth the results here.

examples/rds.txt
```diff
@@ -0,0 +1,111 @@
...
+                "message": "2021-07-23T05:45:54.513995Z\t  199 Query\tSHOW FULL COLUMNS FROM `testdb`.`customers`"
+            }
+        ]
+    }
+
+RDS MYSQL SLOWQUERY LOGS:
+    {
+        "messageType": "DATA_MESSAGE",
+        "owner": "908047316593",
+        "logGroup": "/aws/rds/instance/mysql-db-for-logs/general",
```
dt-be, line 61: Like @wojtasskorcz mentioned [here](https://github.com/dynatrace-oss/dynatrace-aws-log-forwarder/pull/32#discussion_r678916731) isn't this some copy paste error?

image

src/config/rds_mysql.json
mmajcher: Yes I discussed this with @mswiatkowska and we decided that it seems safe.
I still wish I had timing estimates for 10MB and 50MB (log volume per minute).

10k with 3.2s seems fine but we can do 50k max (15s) AND I believe this perf test was done on super fast computer not on little 1vthread lambda so there is still some uncertainty. So 15s might grow considerably but same applies to scenario without this "contains". Guess we'll see :) and we might prioritize perf improvments once again if we see shortcomings in production scenarios

dt-be: With performance you're never sure. In case this makes your mind a little easier, I think my PC is newer/faster than Maria's and when I did the tests both locally and in AWS, AWS was just mildly slower.

examples/rds.txt
mswiatkowska: Yes, it's copy paste error...
image

wojtasskorcz (approved):

dt-be (approved):

tests/unit/logs/test_transformation_full.py
dt-be: Maria confirmed, resolving.

tests/unit/logs/test_transformation_full.py
```diff
@@ -665,6 +665,139 @@
...
+                }
```
dt-be, line 746: Just a note, to have a trace in the future:

For /audit logs here, `audit.user` could be extracted: 
"20210722 17:15:28,ip-10-1-3-230,**admin**,157.25.19.100,13,0,DISCONNECT,testdb,,0,SSL"
For non-audit ones `db.user` could be extracted. 

`db.user` is already extracted for (?) Azure MySql logs, as there it was easy and cheap (separate field) - vs for AWS need to parse the message content. Therefore, this is deliberately not done for AWS (I believe @mswiatkowska discussed this), as it is not cheap enough, and DataProcessingPipelines would be a better target solution for that.

Same goes for `db.name` mentioned above - also filled for Azure, but not in AWS, same reasons:
"message": "20210722 17:15:28,ip-10-1-3-230,admin,157.25.19.100,13,0,DISCONNECT,**testdb**,,0,SSL"

PR #34, opened 2021-07-27
 APM-313824 add exception throttling

equardo:
Still in progress of writing unit tests for this but it seems to work when testing manually. 
Screenshot 2021-07-27 140736
Only 10 exceptions are printed.
image
Exceptions from different spots in code are counted separately.


src/index.py
```diff
@@ -48,17 +48,20 @@ def handler(event, lambda_context):
         result = TransformationResult.Ok
 
     except CallThrottlingException:
-        log_multiline_message("Call Throttling Exception, Kinesis batch will be marked as OK, but some data is dropped")
+        log_multiline_message("Call Throttling Exception, Kinesis batch will be marked as OK, but some data is dropped",
+                              os.path.basename(__file__) + "_" + handler.__name__ + "_1")
```
mmajcher, line 52: I would just put some string constant as 2nd param in all usages. Like "call-throttling-exception" etc
What do you think @dt-be @wojtasskorcz ?

mmajcher: It looks a bit messy this way for me

src/util/logging.py
```diff
@@ -17,30 +17,44 @@
 
 from util.context import Context
 
+log_call_count = dict()
+
+
+def log_multiline_message(message: Text, caller: Text):
+
+    # display logs called from one spot no more than 10 times, "None" means it was counted before
+    if caller != "None":
```
mmajcher, line 26: you can combine this if and nested if with "AND"

src/util/logging.py
```diff
@@ -17,30 +17,44 @@
...
+        if log_call_count.get(caller, 0) < 10:
+            log_call_count[caller] = log_call_count.get(caller, 0) + 1
```
mmajcher, line 28: here, if we are at last possible occurence (lets say limit is 10 and this is 10th call) then I think we should add log entry "Logs for caller X have reached throttling limit of Y, further logs from this caller will be discarded" 

like: 
if calls < 10 - bump dict
if calls == 10 - print warning and return
if calls > 10 - return

src/util/logging.py
```diff
@@ -17,30 +17,44 @@
...
 from util.context import Context
 
+log_call_count = dict()
+
+
+def log_multiline_message(message: Text, caller: Text):
+
+    # display logs called from one spot no more than 10 times, "None" means it was counted before
+    if caller != "None":
+        if log_call_count.get(caller, 0) < 10:
```
mmajcher, line 27: please extract 10 as a top level variable "LOG_THROTTLING_LIMIT_PER_CALLER"

src/util/logging.py
```diff
@@ -17,30 +17,44 @@
...
     if context.debug:
-        log_multiline_message("DEBUG: " + message)
+        log_multiline_message("DEBUG: " + message, caller)
 
 
-def log_error_with_stacktrace(e: Exception, msg):
-    log_multiline_message(f"Exception '{e}' occurred. Additional message: '{msg}'")
-    log_multiline_message(traceback.format_exc())
+def log_error_with_stacktrace(e: Exception, msg, caller: Text):
+    if log_call_count.get(caller, 0) < 10:
```
mmajcher, line 43: Since you are going to log_multiline_message anyway then this checks can and should stay in one place, in log_multiline_message. You can pass caller umodified. 
Since you are calling log_multiline_message twice and we don't want to break the throttling counter you could
log_multiline_message (..... , caller)
log_multiline_message (....., caller+"traceback")

src/util/logging.py
equardo: Then the else would need to be a separate if with two conditions, is that preferable?

mmajcher: You are right, this way is better in such case, my mistake :)

src/util/logging.py
equardo: Maybe I could add an additional parameter to the function, like def log_multiline_message(message: Text, caller: Text, counted: bool = false) and set it to true the second time? We could be sure there would be no overlap with a function name

mmajcher: Both will work fine but I find method signature you proposed a little non-obvious, hard to understand if you don't know this context. If I imagine myself being a person new to this and calling log_multiline_message for some reason, "what the hell is counted = bool ?? is this whether I want to apply throttling counting to my log?". But if you feel that counted=false is better then I am ok with this

src/util/logging.py
```diff
@@ -17,30 +17,53 @@
...
+LOG_THROTTLING_LIMIT_PER_CALLER = 10
+
+log_call_count = dict()
+
+
+def log_multiline_message(message: Text, caller: Text, count_for_throttling: bool = True):
+
+    # display logs called from one spot no more than the specified amount of times,
+    # count_for_throttling means we don't add to the counter but can still throttle if it already exceeds the limit
+    if count_for_throttling is True:
```
mmajcher, line 29: Looking at this now, I am even more convinced that this logic would be much more simpler if we don't introduce count_for_throttling :D

equardo: Wouldn't it be the same thing but instead of count_for_throttling we would check if caller ends with "traceback" also allowing for potential overlap with a function name? This way also seems easier if someone were to add another function calling log_multiline_message multiple times

src/index.py
wojtasskorcz: Hmm, this would require coming up with descriptive strings and I can see potential name clashes. What we need here is a mechanism to generate unique strings easily, and copy-pasting the same line all over the codebase seems easy enough. But I can see your point that it's not the easiest to read. It's a tradeoff between readability and ease of implementation.

src/util/logging.py
equardo: Unless you meant to just count each call separately in the dictionary, I thought about it but it felt inefficient, maybe it is better to keep the function call clean

wojtasskorcz: I think it makes sense to append something to the caller string in the second invocation of `log_multiline_message()` but maybe something more self-descriptive like `"-second-invocation"`

equardo: Also this would mean we would need to check if something is a second invocation when logging that the number of calls was exceeded, so we don't get the message twice.

src/index.py
mmajcher: - If there is a clash we can live with this. 
- The string does not have to be very descriptive, it is not important in any way. 
- It's much easier to put some string than copy this "rule" here and adjust it and import os.

src/util/logging.py
mmajcher: I don't think we need to check anything. One of our internal methods has specific logic that it calls this method twice (even though user only called "log_with_exception" once) and this is a special case and we don't expect any more methods like this in near future.
There is one other simple solution. Merge both invocations from log_with_exception to one invocation. ;)

equardo: Merging them is the cleanest solution but it means the message and traceback need to be in the same line - I'm not sure how much of a problem that is. As for the first part, you mean that it's fine for the logs to read something like this?
image

src/index.py
wojtasskorcz: Alright, I guess clashes are indeed not that bad. We'll throttle after 5 occurrences instead of 10 which is okay. If we're not paying too much attention to make the string unique then I agree that having a custom one may be easier than copying this "rule". Let's change it then

src/util/logging.py
wojtasskorcz: Yeah, let's merge these two invocations. These flattened tracebacks are not readable anyway so if we append them to the end of exception message, readability won't suffer substantially

mmajcher: 😆 yeah they are not readable. 
Just for context:
if you use multiple lines with \n in cloudwatch (normal python behaviour) then you have multiple cloudwatch events, one for each line
ir you substitute \n to \r, then in cloudwatch you have all good in one event but locally you have mess.
Flattening to one line with some separator (what we do) is a (arguably not best) compromise :)

src/index.py
equardo: Good points, change is now pushed

src/util/logging.py
equardo: With current push log_multiline_message is called only once in log_with_exception

src/util/http_client.py
```diff
@@ -11,7 +11,7 @@
 #   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
-
+import os
```
mmajcher, line 14: is this import still needed?

equardo: No, must have missed it

tests/unit/logs/metadata_engine/test_metadata_engine.py
```diff
@@ -55,3 +57,11 @@ def run_with_clean_engine(input):
     output = {}
     engine.apply(input, output)
     return output
+
+
+"""
+@patch.object(MetadataEngine, 'apply')
```
mmajcher, line 63: Do you want co commit this?

mmajcher (dismissed):

tests/unit/logs/metadata_engine/test_metadata_engine.py
equardo: No

tests/unit/logs/metadata_engine/test_metadata_engine.py
```diff
@@ -11,10 +11,12 @@
 #   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
+from unittest.mock import patch
```
wojtasskorcz, line 14: Changes in this file should all be discarded

wojtasskorcz (approved):

mmajcher (dismissed):

wojtasskorcz (dismissed):

mmajcher (dismissed):

src/util/logging.py
```diff
@@ -17,30 +17,47 @@
...
+def log_multiline_message(message: Text, caller: Text):
+
+    # display logs called from one spot no more than the specified amount of times
+    if log_call_count.get(caller, 0) < LOG_THROTTLING_LIMIT_PER_CALLER:
+        log_call_count[caller] = log_call_count.get(caller, 0) + 1
+    elif log_call_count.get(caller, 0) == LOG_THROTTLING_LIMIT_PER_CALLER:
+        log_call_count[caller] = log_call_count.get(caller, 0) + 1
+        print(f"Logging calls from caller '{caller}' exceeded the throttling limit of",
+              f"{LOG_THROTTLING_LIMIT_PER_CALLER}. Further logs from this caller will be discarded")
+        return
```
dt-be, line 34: This is a bit verbose would be a bit unreadable if e.g. I entered here just to see how messages are logged, I would have to read more into it than I would like to. I can give you an example of what I think might be better.

Btw of missing things because it's too verbose - is there a print missing in the first if, or did I just get lost?

src/util/logging.py
```diff
@@ -17,30 +17,47 @@
...
+    else:
+        return
```
dt-be, line 36: This is a bit verbose would be a bit unreadable if e.g. I entered here just to see how messages are logged, I would have to read more into it than I would like to. I can give you an example of what I think might be better.

Btw of missing things because it's too verbose - is there a print missing in the first if, or did I just get lost?
Ok, I see it's there later - but I still got lost because of the verbosity of the first check. I'd extract the whole of those if to a separate method. It could be done directly (I recommend PyCharm Extract method if you don't know it), or here's my draft take on it, just to show you what I have in mind:
```
if check_log_limit_exceeded:
	return
```	

```	
def check_log_limit_exceeded(caller):
	log_limit_left = log_call_count.get(caller, 0) - LOG_THROTTLING_LIMIT_PER_CALLER
	if log_limit_left == 1:
           print(f"Logging calls from caller '{caller}' exceeded the throttling limit of",
              f"{LOG_THROTTLING_LIMIT_PER_CALLER}. Further logs from this caller will be discarded")
	bool is_loggable = log_limit_left == 0
	return is_loggable
```

equardo: Done, thank you

src/util/logging.py
```diff
@@ -17,30 +17,54 @@
...
+    log_calls_left = LOG_THROTTLING_LIMIT_PER_CALLER - log_call_count.get(caller, 0)
+
+    if log_calls_left == 0:
+        log_call_count[caller] = log_call_count.get(caller, 0) + 1
+        print(f"Logging calls from caller '{caller}' exceeded the throttling limit of",
+              f"{LOG_THROTTLING_LIMIT_PER_CALLER}. Further logs from this caller will be discarded")
+
+    caller_exceeded_limit = log_calls_left <= 0
+    if not caller_exceeded_limit:
+        log_call_count[caller] = log_call_count.get(caller, 0) + 1
```
dt-be, line 45: It looks much better, thanks.

For log_calls_left == 0 though, aren't you bumping up the number twice (39 & 46)? That would still work but seems wrong. I think I'd just remove 39 and it would be nice and clean.
`if log_calls_left == 0:` > true

`caller_exceeded_limit = log_calls_left <= 0
    if not caller_exceeded_limit:` > true

equardo: Unless I am missing something, when log_calls_left == 0, it goes into the first if, but caller_exceeded_limit evaluates to True and doesn't go into the second if

dt-be: Looked at it too long, I missed `not` xD
Sorry, good to go:P

dt-be (dismissed):

mmajcher (approved):

wojtasskorcz (approved):

mmajcher (dismissed):

mmajcher (approved):

PR #35, opened 2021-07-28
APM-311221 - Added parser for AWS SNS logs

bmrozinski-dt:
- added config for SNS
- added SNS cases to transformation test


src/config/sns.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "SNS",
+  "displayName": "Simple Notification Service",
+  "rules": [
+    {
+      "aws": {
+        "logGroup": "sns/%{DATA:region_id}/%{DATA:account_id}/%{USERNAME:resource_id}%{GREEDYDATA:failure_suffix}"
```
dt-be, line 7: **/**%{USERNAME:resource_id}%{GREEDYDATA:failure_suffix}
doesn't it need / between those two as well? 
Not 100% sure how it works, asking.

"logGroup": "sns/us-east-1/444000444/sample-sns-logs-generator/Failure",
"logGroup": "sns/us-east-1/444000444/sample-sns-logs-generator",

mmajcher, line 7: I think there is a test for grok matching on log group. Please add multiple testcases for this specific pattern so we can see more clearly how this matching with USERNAME works :) I am not sure how this works

mmajcher: I think you can't have this slash put here as it is optional. I think username will match all until slash or end. It if matched until slash then "failure_suffix" will contain "/Failure"

src/config/sns.json
```diff
@@ -0,0 +1,41 @@
...
+        },
+        {
+          "key": "aws.arn",
+          "priority": 1,
+          "pattern": "format_arn('arn:{}:sns:{}:{}:{}', [partition, region, account_id, resource_id])"
+        },
+        {
+          "key": "severity",
+          "pattern": "if( failure_suffix == '', &'INFO', &'ERROR', @)"
+        },
```
dt-be, line 33: I thought we should get audit.result for those error logs, similiarly as for CloudTrail.
`\\\"ErrorCode\\\":\\\"AccessDenied\\\`
```
        {
          "key": "audit.result",
          "description": "If a field errorCode exists, then return Failure.{errorCode}, else return Succeeded",
          "pattern": "if( log_content.errorCode==null, &'Succeeded', &[`Failed`, log_content.errorCode] | [? @!=null ] | join('.', @) , @)"
        },
```
You would also need to add
`  "logContentParseAs": "json"`
for that, so I expect resistance here 😄 and PawelS is not here for me to ask and make sure when do we want this field filled and where not.

mmajcher: From what I know we don't need this right now (use case is following: you select all logs with "ERROR" and then look at them on your own, that's all we need). But if you see some use cases or expect Pawel to see some, can you follow up this with Pawel?

mmajcher (approved):

wojtasskorcz (approved):

src/config/sns.json
dt-be: Followed up with Marcin, final decision was this should not be done, but not because "we don't need this right now". The actual reason was to keep "**audit.XXX**" fields reserved for "audit" logs, i**.e. logs about configuration/deployment changes**, and this one here seems to be "message accepted/or not" rather than "SNS settings changed successfully or not".

PR #36, opened 2021-07-28
APM-314438: Redesign VPC to allow SFM push from Lambda; restore SFM

mmajcher:


src/logs/self_monitoring/sfm.py
```diff
@@ -195,3 +206,17 @@ def _prepare_cloudwatch_metric(metric_name, value: Union[int, float, list], unit
         cw_metric["Value"] = value
 
     return cw_metric
+
+
+def _prepare_cloudwatch_metric_statistic(metric_name, unit, dimensions, minimum, maximum, sum, count) -> dict:
```
wojtasskorcz, line 211: Let's maybe make `_prepare_cloudwatch_metric` consistent with this method and take value as the last argument?

wojtasskorcz (dismissed):

dynatrace-aws-log-forwarder-template.yaml
```diff
@@ -649,6 +658,20 @@ Resources:
...
+      Domain: VPC
+
+  VPCNatGateway:
+    Condition: DeployAGwithVPC
+    Type: "AWS::EC2::NatGateway"
+    DependsOn: VPCNatGatewayElasticIP
+    Properties:
+      AllocationId: !GetAtt VPCNatGatewayElasticIP.AllocationId
+      SubnetId: !Ref VPCPublicSubnet1
+
```
dt-be, line 674: Mostly a note to me, please check if I read this right:
*AG EC2 (:621) & Lambda(:83) now run in the private subnet of the VPC.
*private(:635) and public(:644) Subnets are in same VPC
*NATGateway is in Public subnet (673), with an elastic IP (672)
*all traffic(708) from private subnet goes through NATGateway(709) in public (via routing table)

I'm wondering how is NATGateway from Public subnet visible in the Private one, but I assume that there is some logic/magic that happens behind the line 709 that NatGatewayId, since AWS has a dedicated field for that.

dynatrace-aws-logs.sh
```diff
@@ -242,7 +242,7 @@ EOF
 
   set -e
 
-  echo "Deploying stack $STACK_NAME"
+  echo "Deploying stack $STACK_NAME. This might take about 10 minutes if you chose included AG option, or about 5 minutes if you chose existing AG option."
```
dt-be, line 245: ? Is this a consequence of changes with subnets/VPCs in this PR? Because that would seem like a penalty on existingAG setup for some improvement for the other (newAG) one and to me at least a warning sign to reconsider what we're doing. I never noticed it taking 5 min so far, but certainly I wasn't looking, so asking.

dynatrace-aws-log-forwarder-template.yaml
```diff
@@ -637,6 +637,15 @@ Resources:
       CidrBlock: 172.31.1.0/27
       MapPublicIpOnLaunch: true
 
+  VPCPrivateSubnet1:
```
dt-be, line 640: Super minor, but is there any reason for those 1s at the ends?

tests/unit/logs/self_monitoring/test_sfm.py
```diff
@@ -50,19 +52,19 @@ def test_self_monitoring_context():
             'Dimensions': [{'Name': 'function_name', 'Value': 'my-lambda-function'}],
             'MetricName': 'Kinesis record age',
             'Unit': 'Seconds',
-            'Values': [5]
+            'StatisticValues': {'Maximum': 10, 'Minimum': 5, 'SampleCount': 2, 'Sum': 15},
```
dt-be, line 55: Sum seems worthless:P I assume it's required to report those 4, and Avg is calculated from it. 

Does our dashboard show something reasonable after those changes though? Again, if it's sum there, it's worthless, if avg would be okayish, although I'd still add min+max there now that we have them instead.

dynatrace-aws-logs.sh
equardo: I don't feel it ever took 10 min with included AG either. I also believe the option is referred to as "new AG" in the docs most of the time. One more minor thing is I think I would say "This will take about.." since you still have uncertainty by saying "about" - I don't see what additional uncertainty the "might" would refer to.

src/logs/self_monitoring/sfm.py
```diff
@@ -102,12 +102,23 @@ def _generate_metrics(self):
...
-        metrics.append(_prepare_cloudwatch_metric(
-            "Kinesis record age", self._kinesis_records_age, "Seconds", common_dimensions))
-        metrics.append(_prepare_cloudwatch_metric(
-            "Kinesis record.data compressed size", self._record_data_compressed_size, "Bytes", common_dimensions))
-        metrics.append(_prepare_cloudwatch_metric(
-            "Kinesis record.data decompressed size", self._record_data_decompressed_size, "Bytes", common_dimensions))
+        metrics.append(_prepare_cloudwatch_metric_statistic(
+            "Kinesis record age", "Seconds", common_dimensions,
+            min(self._kinesis_records_age), max(self._kinesis_records_age),
+            sum(self._kinesis_records_age), len(self._kinesis_records_age)
```
dt-be, line 108: min, max, sum, len would fit better inside the method imho, with just the proper metric (self._kinesis_records_age) being passed

dynatrace-aws-logs.sh
dt-be: For me it's about it *might* take less:P Good point about the naming.

src/logs/self_monitoring/sfm.py
wojtasskorcz: Great idea!

src/logs/self_monitoring/sfm.py
mmajcher: Fixed

dynatrace-aws-logs.sh
mmajcher: No, nothing changes with regard to time (maybe a little, new subnet and NatGateway).
I just checked and this deployment does take 5 minutes for Cloudformation (with included AG). 

I just pushed an update, please check if you like it. I put "up to 10 minutes" without separating scenarios. This has only one goal - if user sees "waiting for stack update/create" for 2nd minute, he does not panic but sees this info and keeps waiting. It does not hurt to give longer estimate here, I think AWS can also surprise sometimes and have its blips. 

For the naming, we are rebranding this in the docs right now (as discussed within team some time ago). There will be includedAG and existingAG. It's 95% sure at the moment, I am consulting this right now with Jarek Orlowski. He proposed "CloudFormation-included" and also "stack-included" etc.

dynatrace-aws-log-forwarder-template.yaml
mmajcher: No reason, just removed numbering 👍

tests/unit/logs/self_monitoring/test_sfm.py
mmajcher: This is the API of this method. Avg is calculated from it. 
Our dashboard works well with this as is! (I was surprised but multi-value push and statistical-value push seem to resolve to same thing underneath) In the dashboard we have min, max, avg

src/logs/self_monitoring/sfm.py
mmajcher: You are right, much cleaner, fixed

dynatrace-aws-logs.sh
mmajcher: Ok, talked with Jarek and we settled for stack-included after language consultations with Malcolm

wojtasskorcz (approved):

PR #37, opened 2021-08-10
APM-316012 aurora RDS

karolkieglerski:
without severity profiling:

- 2.603 seconds
- 2.472 seconds
- 2.481 seconds

with that profiling:

- 3.397 seconds
- 3.198 seconds
- 3.234 seconds


src/config/rds_aurora.json
```diff
@@ -0,0 +1,50 @@
...
+          "key": "dt.source_entity",
+          "pattern": "dt_meid_rds_v2(\"aws.arn\")"
+        },
+        {
+          "key": "severity",
+          "pattern": "if(log_type == 'slowquery' || contains(log_content, '[Warning]'), &'WARNING', &if(contains(log_content, '[Error]'), &'ERROR', &'INFO', @), @)"
+        },
+        {
+          "key": "log.source",
+          "pattern": "join( ' - ', ['rds', join(' ', [log_type, 'logs']) ])"
```
mmajcher, line 45: I think we have function "format" that would be more readable here

src/config/rds_aurora.json
```diff
@@ -0,0 +1,50 @@
...
+          "pattern": "format_arn('arn:{}:rds:{}:{}:db:{}', [partition, region, account_id, \"aws.cluster.instance\"])",
+          "priority": 10
+        },
+        {
+          "key": "dt.source_entity",
+          "pattern": "dt_meid_rds_v2(\"aws.arn\")"
+        },
+        {
+          "key": "severity",
+          "pattern": "if(log_type == 'slowquery' || contains(log_content, '[Warning]'), &'WARNING', &if(contains(log_content, '[Error]'), &'ERROR', &'INFO', @), @)"
```
mmajcher, line 41: This is tricky ground, let's assume we only have INFO logs then for each log we do 2 times check-content-run-on-whole-string. I am not sure if this should go inside without some basic perf test @mswiatkowska knows how to do them ;)
But I am open to other opinions, if you feel safe with this (remember we need to be able to do 1-10MB in minute) then fine

mmajcher: I think this is time to create some automated perf test. But this is also tricky since CPUs are not made equal! ;)

mswiatkowska: This is exactly the same pattern I've used for other rds databases. 
I'm only wondering - we will put proper Warning/Error severity for mysql slowquery and warning logs (and other logs containing [Error]/[Warning]) but e.g. postgres can produce some error logs like: 
`2021-08-09 10:08:56 UTC:10.0.0.123(52834):postgres@tstdb:[20175]:ERROR: column "wrong" does not exist at character 8`
and then we will classify this log as INFO.
Is it ok with us? Or maybe we want to just remove this level check as we are not able to classify all logs properly?

karolkieglerski: thats true, in postgres severity is in `:` instead of `[]` ex. `:LOG:`

karolkieglerski: we must decide if we lis add for postgresql or remove for all

karolkieglerski: I splitted as we decided

src/config/rds_aurora.json
```diff
@@ -0,0 +1,56 @@
...
+          "pattern": "dt_meid_rds_v2(\"aws.arn\")"
+        },
+        {
+          "key": "severity",
+          "pattern": "if(log_type == 'slowquery', &'WARNING', &null, @)",
+          "priority": 11
+        },
+        {
+          "key": "severity",
+          "pattern": "if(log_type == 'postgresql', &if(contains(log_content, ':WARNING:'), &'WARNING', &if(contains(log_content, ':ERROR:'), &'ERROR', &'INFO', @), @), &if(contains(log_content, '[Warning]'), &'WARNING', &if(contains(log_content, '[Error]'), &'ERROR', &null, @), @), @)",
```
karolkieglerski, line 46: for all 10 types of pytest cases I tun it's multiply by 1000 so 10 000 requests and got:
```
24930185 function calls (23957927 primitive calls) in 27.175 seconds
```

mmajcher: I am not sure I understand. How much data has been transformed in 27seconds? 1MB? 5MB?

mmajcher, line 46: Can we split this entry into 2? One with if postgresql then something (otherwise null), second one if mysql then something (otherwise null).
And then 3d one if severity null then severity=INFO

mmajcher: Is this only for this specific jmespath expression or whole config? Please add more context

karolkieglerski: If was before but @mswiatkowska found another use case that it can't

karolkieglerski: 10 000 test cases as we have right now for all databases processed by pytest

karolkieglerski: from this line https://github.com/dynatrace-oss/dynatrace-aws-log-forwarder/blob/7320460fad726b06abfa15fdc96c5a0b101722f1/tests/unit/logs/test_transformation_full.py#L668 to the bottom multiply 1000 times

karolkieglerski: but I don't have trust to that perf library right now without that additional section it took 26s but with that severity parsing it took 31s

mmajcher: This numbers tell me nothing. Perf for this should say following:
I pushed X MB of log entries (number of entries is Y) through this specific json config and it took Z seconds.
I don't care how fast all of our testcases are

mswiatkowska: If we wanted to have 3rd separate 'if' we would need to add additional check for if not slowquery and not postgresql -> it's too much really :D

mswiatkowska: I really think about removing this severity conditions and send INFO for all rds logs...

karolkieglerski: @mmajcher ok, so where we have performance test that I can check how the performance was degraded or not?

mmajcher: I don't get it, 3rd would be if mysql then ...

mmajcher: The performance surely is degraded but question is how much
Unfortunately we don't have one so if you with @mswiatkowska decide that you feel safe then push as is :)

mmajcher: Ok, I get it now. I wish json had comments so we could add this info

karolkieglerski: merged your branch and move to separate stage

src/requirements-dev.txt
```diff
@@ -1,3 +1,4 @@
 pytest==6.2.2
+pytest-profiling==1.7.0
```
mmajcher, line 2: Is this used anywhere?

mmajcher (approved):

src/requirements-dev.txt
karolkieglerski: https://github.com/dynatrace-oss/dynatrace-aws-log-forwarder/pull/37/files#diff-6ac3f79fc25d95cd1e3d51da53a4b21b939437392578a35ae8cd6d5366ca5485R30

karolkieglerski: or you can use it locally to measure time of your tests

mswiatkowska (dismissed):

.travis.yml
```diff
@@ -20,13 +20,14 @@ stages:
 jobs:
   include:
     - stage: unit_tests
-      name: Execute unit tests & linting
+      name: Execute unit & pref tests + linting
```
mswiatkowska, line 23: typo `pref`

mswiatkowska (approved):

Dawid-Kaszubski (approved):

karolkieglerski:
@mmajcher can you review?

karolkieglerski:
@mmajcher or @dt-be can you look at this PR?

mmajcher (approved):

src/config/rds_aurora.json
dt-be: You can always do it like that - I add description/whatever field that will get discarded when rules are loaded into class -  imho worth it for confounded ones.
```
        {
          "key": "audit.result",
          "description": "If a field errorCode exists, then return Failure.{errorCode}, else return Succeeded",
          "pattern": "if( log_content.errorCode==null, &'Succeeded', &[`Failed`, log_content.errorCode] | [? @!=null ] | join('.', @) , @)"
        },
```

tests/perf/logs/test_transformation_full.py
```diff
@@ -0,0 +1,126 @@
...
+            'aws.region': "us-east-1",
+            'aws.account.id': "444000444",
+            'severity': 'INFO',
+            'log.source': 'rds - general logs',
+        },
+        "perf_check": {
+            "repeat_record": 10000,
+            "time_limit_sec": 5
+            # roughly 1,5MB of content, ~7MB counting whole log entry with metadata
+        }
```
dt-be, line 86: Really nice to have this test, appreciated.

src/config/rds_aurora.json
karolkieglerski: @dt-be it was already merge before I saw your comment, but will prepare another PR with that description

PR #39, opened 2021-08-11
Noissue: Aurora perf test

mmajcher:
I wrote quick perf test for you.
Perf test result is very good 👍 see the comment

But this is on my very fast computer. Also there might be some optimizations going on underneath. So this is not very certain but it's best what we have at the moment.

I would love to run this with randomized log content and in lambda environment with their basic vCPU.

Can you please organise this and include into codebase? I think this should be in some separate test. Otherwise this test will get very long once we add more check. Not sure how it should look like. But if transformation_full.py takes too long this might discourage devs from this test :D

### DIFF
--- a/tests/unit/logs/test_transformation_full.py
+++ b/tests/unit/logs/test_transformation_full.py
@@ -13,6 +13,7 @@
 #   limitations under the License.
 
 import json
+import time
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


mmajcher:
I guess cleanup here would be to move this to maybe tests/perf by copying this specific test but with only chosen pytest param(s). So then you can run tests/unit and tests/perf separately locally. In travis job we would have to add tests/perf execution

karolkieglerski (approved):

PR #41, opened 2021-08-25
description added

karolkieglerski:


dtwwolinski (approved):

src/config/rds.json
```diff
@@ -38,6 +38,7 @@
         },
         {
           "key": "severity",
+          "description": "If a field log_type equal postgresql, then (If a field log_content contains :WARNING:, then returns WARNING, else (If log_content contains :ERROR:, then returns ERROR, else returns INFO)), else (If a field log_content contains [Warning], then returns WARNING, else (If log_content contains [Error], then returns ERROR, else returns default INFO))",
```
mmajcher, line 41: This should be like 
"For postgres we check for ':WARNING:', ':ERROR:' (otherwise we give INFO), otherwise we treat is as mariadb (?) and there we check for YYY ZZZZ "

As is this is simply a rewrite of what's below and not very useful. We need simple explanation like "you don't have to read this all, you need to know that we handle postgres and if we don't match postgres then we treat it as mariadb"

PR #43, opened 2021-09-15
APM-306046: Restore LogGroup sfm using AWS native metrics

mmajcher:


dynatrace-aws-log-forwarder-template.yaml
```diff
@@ -297,14 +297,14 @@ Resources:
               "type": "metric",
               "properties": {
                 "metrics": [
-                [ { "expression": "SEARCH('{DT/LogsStreaming,function_name,log_group} function_name=\"${__LambdaName__}\" MetricName=\"Log entries by LogGroup\"', 'Sum', 60)", "label": "Log entries - [${!PROP('Dim.log_group')}]", "id": "e1", "region": "${__Region__}" } ]
+                [ { "expression": "SEARCH('{AWS/Logs,FilterName,LogGroupName,DestinationType} FilterName=\"${__StackName__}\" MetricName=\"ForwardedLogEvents\"', 'Sum', 60)", "label": "[${!PROP('Dim.LogGroupName')}]", "id": "e1", "region": "${__Region__}" } ]
```
wojtasskorcz, line 300: Even though we don't really want to see splitting by `DestinationType`, we have to list it here because CloudWatch won't automatically aggregate over it. Do I remember correctly?

wojtasskorcz (approved):
Do you perhaps have any screenshots to show how the end result looks like?

dynatrace-aws-log-forwarder-template.yaml
mmajcher: I tried without it and it does not work. Not even aggregate, the search does not find it I think. 
I think that for AWS to find specific metric you need a combination of all dimensions. Otherwise it is a different metric. I am not sure how to explain it better but that's how I understand it

wojtasskorcz: Yes, that's how I remembered it and what I had in mind. You can't omit a dimensions and expect AWS to aggregate over it -- it'll just return an empty result. Wanted to make sure

mmajcher:
Screenshot 2021-09-14 135437
This is how it looks like for a simple setup. It is exactly same as it was working before.
This orange log group is only a point because I sent only one log for testing ;) But this was same before. Only source changed

dt-be (approved):

PR #50, opened 2021-12-06
feat: Adding log retention

edogden-pbx:
The following change adds log retention to the Lambda functions LogGroup.
This should help reduce the amount of Lambda logs stored in AWS which will cut down on CloudWatch log storage costs.

In this propose I have defaulted to 90 days as a good baseline, however we might want to reduce this further.


edogden-pbx:
The only issue I have come across is that if you are updating an existing stack your CloudFroamtion will fail because the LogGroup already exists.

edogden-pbx:
Found a better way of doing it

PR #51, opened 2021-12-06
feature: Adding export name

edogden-pbx:
This feature change adds export names to the `FirehoseArn` and `CloudWatchLogsRoleArn` output values.
This is done so you could dynamically import the values to a LogSubscriptions via a [Function ImportValue](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-importvalue.html).

I chose a very generic name of AWS::StackName:(FirehoseARN/CloudWatchARN) however this is up to debate.

### DIFF
--- a/dynatrace-aws-log-forwarder-template.yaml
+++ b/dynatrace-aws-log-forwarder-template.yaml
@@ -746,10 +746,14 @@ Outputs:
   FirehoseArn:
     Description: "Firehose ARN"
     Value: !GetAtt FirehoseLogStreams.Arn
+    Export:
+      Name: !Join [ ":", [ !Ref "AWS::StackName", FirehoseARN ] ]
 
   CloudWatchLogsRoleArn:
     Description: "CloudWatch Logs role ARN allowing streaming to Firehose"
     Value: !GetAtt CloudWatchLogsRole.Arn
+    Export:
+      Name: !Join [ ":", [ !Ref "AWS::StackName", CloudWatchARN ] ]
 
   EC2ActiveGateHostname:
     Description: "EC2ActiveGateHostname"


dynatrace-aws-log-forwarder-template.yaml
```diff
@@ -746,10 +746,14 @@ Outputs:
...
     Description: "Firehose ARN"
     Value: !GetAtt FirehoseLogStreams.Arn
+    Export:
+      Name: !Join [ ":", [ !Ref "AWS::StackName", FirehoseARN ] ]
 
   CloudWatchLogsRoleArn:
     Description: "CloudWatch Logs role ARN allowing streaming to Firehose"
     Value: !GetAtt CloudWatchLogsRole.Arn
+    Export:
+      Name: !Join [ ":", [ !Ref "AWS::StackName", CloudWatchARN ] ]
```
dt-be, line 756: Could you tell me more about your use case? The way I understand it, you are using `deploy` option, but use something else than our `subscribe` for subscribing, and would like to get those outputs available for use there (i.e. some other CloudFormation stack that deploys the actual resources that produce logs?), without actually querying cli for those arns the way `subscribe` does it.

edogden-pbx: @dt-be the use-case we have for an export name is for codifying the output values in our serverless config. Rather than running the script to subscribe the LogGroups we can add the `serverless-plugin-log-subscription` plugin to our serverless lambda that can import the Cloudwatch and Firehouse ARNs directly from the stack and auto create a LogSubscription to attach to the lambda LogGroups

Example serverless.yml code
```yaml
service: my-service

plugins:
  - serverless-plugin-log-subscription

custom:
  logSubscription:
    - enabled: true
      destinationArn: 'Fn::ImportValue': 'dynatrace-aws-logs-FirehouseARN'
      roleArn: 'Fn::ImportValue': 'dynatrace-aws-logs-CloudWatchARN'
```

dt-be:
@edogden-pbx Same as in the other PR, please merge dt-oss master to your branch here

dt-be (approved):

kamiljsokolowski (approved):

PR #53, opened 2021-12-07
feature: Adding log retention

edogden-pbx:
The following feature change adds log retention to the Lambda functions LogGroup.
This should help reduce the amount of Lambda logs stored in AWS which will cut down on CloudWatch log storage costs.

In this propose I have defaulted to 90 days as a good baseline, however we might want to reduce this further.

As found per my previous PR https://github.com/dynatrace-oss/dynatrace-aws-log-forwarder/pull/50 this is set in the run script to support the updating of the stack.

### DIFF
--- a/dynatrace-aws-logs.sh
+++ b/dynatrace-aws-logs.sh
@@ -259,6 +259,12 @@ EOF
   aws lambda update-function-code --function-name "$LAMBDA_ARN" --zip-file fileb://"$LAMBDA_ZIP_NAME" > /dev/null
   echo; echo "Updated Lambda code of $LAMBDA_ARN"; echo
 
+  LAMBDA_NAME=$(aws cloudformation describe-stack-resource --stack-name "$STACK_NAME" \
+     --logical-resource-id Lambda --query "StackResourceDetail.PhysicalResourceId" --output text)
+    
+  aws logs put-retention-policy --log-group-name "/aws/lambda/$LAMBDA_NAME" --retention-in-days 90
+  echo; echo "Adding log retention policy of 90 days"; echo
+
   # SHOW OUTPUTS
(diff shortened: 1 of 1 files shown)


dt-be:
Seems a reasonable thing to add. 
I'd much rather see it as a part of cloudformation than as a separate call though - there are a lot of things configured there already and I would keep it all there. Something like below maybe, I think it would also work on updating the stack, but please check.

dynatrace-aws-log-forwarder-template.yaml
```
  Lambda:
     ...

  LambdaSelfMonitoringLogGroup:
    Type: AWS::Logs::LogGroup
    Properties:
      LogGroupName: !Sub "/aws/lambda/${Lambda}"
      RetentionInDays: 90
      
  LambdaRole:
    ...
```
      
90 might even be too much, but unless anyone else protests, let's have it for a start.

mmajcher:
Seems good 👍 
I would probably give max 30 days, what do you think?
Otherwise looks good.

Our intention for Contributing is to have external fork merged to a feature branch (not master), we will update you how to proceed.
Because our tests do not run from external fork for security reasons so you will not pass checks. We need to organise this somehow :)

wojtasskorcz (approved):

edogden-pbx:
> Seems good 👍 I would probably give max 30 days, what do you think? Otherwise looks good.
> 
> Our intention for Contributing is to have external fork merged to a feature branch (not master), we will update you how to proceed. Because our tests do not run from external fork for security reasons so you will not pass checks. We need to organise this somehow :)

let me know if you need me to do anything

dt-be:
Hi. 
I've created a target branch for you to merge into, and updated it in the PR.
Please merge current dynatrace-oss/master into your branch, it should be able to merge after that.

kamiljsokolowski (approved):

PR #55, opened 2021-12-08
APM-340873

kamiljsokolowski:
This PR will fix:
1. Disable builds from external forks

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 dist: focal
+if: fork = false
 env:
   global:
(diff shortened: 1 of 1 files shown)


kamiljsokolowski:
@karolkieglerski first of all thanks for the tip with handling forks from pipeline code.

I do however have a concern when it comes to other forwarder build pipelines (GCP and Azure). I can see that you are not using this condition globally at every stage. I think we should unify our approach here.

karolkieglerski (approved):

wojtasskorcz (approved):

PR #61, opened 2022-01-17
APM-318151 Added [TRUNCATED] to log msgs that are being truncated

IrynaKudlaienko:


wojtasskorcz (approved):

src/logs/logs_sender.py
```diff
@@ -147,7 +148,8 @@ def ensure_content_length(log_entry, context):
     log_content_len = len(log_content)
     if log_content_len > DYNATRACE_LOG_INGEST_CONTENT_MAX_LENGTH:
         context.sfm.log_content_trimmed()
-        log_entry["content"] = log_content[0: DYNATRACE_LOG_INGEST_CONTENT_MAX_LENGTH]
+        trimmed_len = DYNATRACE_LOG_INGEST_CONTENT_MAX_LENGTH - len(DYNATRACE_LOG_INGEST_CONTENT_MARK_TRIMMED)
```
bmrozinski-dt, line 151: I don't know if Python would somehow optimize it under the hood (to not calculate `trimmed_len` every time). I suggest simply creating a `DYNATRACE_LOG_INGEST_CONTENT_TRIMMED_LENGTH` constant.

tests/unit/logs/test_logs_sender.py
```diff
@@ -72,6 +72,8 @@ def test_request_and_content_length(self):
 
         logs_sender.DYNATRACE_LOG_INGEST_CONTENT_MAX_LENGTH = 50
         logs_sender.DYNATRACE_LOG_INGEST_REQUEST_MAX_SIZE = 115
+        logs_sender.DYNATRACE_LOG_INGEST_CONTENT_MARK_TRIMMED = "[TRUNCATED]"
```
bmrozinski-dt, line 75: Is this necessary? The value is the same as the one in `logs_sender.py`.

src/logs/logs_sender.py
```diff
@@ -23,10 +23,11 @@
 
 LOGS_API_PATH = "/api/v2/logs/ingest"
 
+DYNATRACE_LOG_INGEST_CONTENT_MARK_TRIMMED = "[TRUNCATED]"
 DYNATRACE_LOG_INGEST_CONTENT_MAX_LENGTH = 8192
 DYNATRACE_LOG_INGEST_ATTRIBUTE_MAX_LENGTH = 250
-DYNATRACE_LOG_INGEST_REQUEST_MAX_SIZE = 1048576 # 1MB
-DYNATRACE_LOG_INGEST_MAX_RECORD_AGE = 86340 # 1 day
+DYNATRACE_LOG_INGEST_REQUEST_MAX_SIZE = 1048576  # 1MB
+DYNATRACE_LOG_INGEST_MAX_RECORD_AGE = 86340  # 1 day
```
bmrozinski-dt, line 30: I was wondering what is this change about, but I assume it was autoformatted. I found this in Python docs:
_An inline comment is a comment on the same line as a statement. Inline comments should be separated by at least two spaces from the statement. They should start with a # and a single space._
https://www.python.org/dev/peps/pep-0008/#inline-comments

IrynaKudlaienko: Yes, that's the reason I left this change. PEP 8: E261 at least two spaces before inline comment. Can I leave it?

src/logs/logs_sender.py
IrynaKudlaienko: I'll take a look.

src/logs/logs_sender.py
bmrozinski-dt: Sure. It's better to follow standards 🙂

tests/unit/logs/test_logs_sender.py
bmrozinski-dt: I don't know if I get the point, but in my opinion, it is absolutely OK for the test to break when changes in source code happen.

wojtasskorcz: I agree with @bmrozinski-dt 
I didn't notice this at first. We definitely want to be informed through tests what functionality will change/break if we change something in the production code. So if we change it in `logs_sender.py` to `[TRUNCATED BY DYNATRACE]`, we want this test to fail and we'll adjust it accordingly

src/logs/logs_sender.py
wojtasskorcz: This would only be calculated for the very long logs (so I assume quite rarely). Moreover, the `len()` function takes constant time, so it's basically just two operations. I think compute time optimization is not really important here. We can add a constant though, just to make it clearer to the person reading the code, that it always computes to the same value

IrynaKudlaienko: As we discussed, I'll leave current implementation. The time complexity of len(string) is constant O(1), and it's a simple math operation.

wojtasskorcz (approved):

wojtasskorcz:
@IrynaKudlaienko delete the branch after merge please :)

PR #64, opened 2022-02-08
Merge external contributions - adding log retention

kamiljsokolowski:


karolkieglerski (approved):

edogden-pbx (approved):

kamiljsokolowski:
Closing this pull request due to the deployment script being incapable of determining Lambda log group name before the function starts logging.

After doing some research it came to attention that the randomly named function log group is being created automagically after the function receives the data from Kinesis. In order to incorporate function log retention, we would need to modify the deployment script AND the CloudFormation template. We would need to create the log group as part of the Lambda deployment and then be able to configure log retention.

As this is a nice-to-have rather than a must-have, we will not be implementing those changes at this moment but we will add them to our improvements list as a potential future update.

PR #69, opened 2022-04-14
Bugfix/apm 363970 aws log forwarder cfn lint issues

equardo:
Remove a superfluous dependsOn - it's already enforced by a GetAtt. 

Remove tags that are no longer allowed for certain resources and caused included AG deployment to fail.

### DIFF
--- a/dynatrace-aws-log-forwarder-template.yaml
+++ b/dynatrace-aws-log-forwarder-template.yaml
@@ -688,7 +688,7 @@ Resources:
     Condition: DeployAGwithVPC
     Type: AWS::EC2::EIP
     Properties:
-      Domain: vpc
+      Domain: VPC
       Tags:
         - Key: Name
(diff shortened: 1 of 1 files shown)


dynatrace-aws-log-forwarder-template.yaml
```diff
@@ -699,7 +696,6 @@ Resources:
   VPCNatGateway:
     Condition: DeployAGwithVPC
     Type: "AWS::EC2::NatGateway"
-    DependsOn: VPCNatGatewayElasticIP
```
wojtasskorcz, line 702: Why can we remove it now?

equardo: !GetAtt in the next lines already implies a dependency on VPCNatGatewayElasticIP

wojtasskorcz: Ahh, right, so it was superfluous previously already, not only after your changes in this PR

wojtasskorcz (approved):

PR #71, opened 2022-07-06
APM-377837: Use existing AG parameter as optional, with TRUE as default

joaquinfilipic-dynatrace:


dynatrace-aws-logs.sh
```diff
@@ -56,7 +56,7 @@ case $MAIN_OPTION in
 
   function print_help_deploy {
     printf "
-usage: dynatrace-aws-logs.sh deploy --target-url TARGET_URL --target-api-token TARGET_API_TOKEN --use-existing-active-gate {true|false} [--target-paas-token TARGET_PAAS_TOKEN] [--require-valid-certificate {true|false}] [--stack-name STACK_NAME]
+usage: dynatrace-aws-logs.sh deploy --target-url TARGET_URL --target-api-token TARGET_API_TOKEN [--use-existing-active-gate {true|false}] [--target-paas-token TARGET_PAAS_TOKEN] [--require-valid-certificate {true|false}] [--stack-name STACK_NAME]
 
 arguments:
   -h, --help            show this help message and exit
```
equardo, line 62: Can't comment the correct line - in 73 we should probably mention that it defaults

dynatrace-aws-logs.sh
```diff
@@ -204,7 +204,7 @@ EOF
   if [ -z "$TARGET_API_TOKEN" ]; then echo "No --target-api-token"; print_help_deploy; exit 1; fi
```
equardo, line 204: Again, here I mean 169 - wouldn't that fail the deployment with no param given?

joaquinfilipic-dynatrace: Exactly, those checks are done in case the argument is present in the first place. There's a case for that, line 154

equardo: I see, thanks! I thought it would be something like that as it would be hard to miss when testing :P

dynatrace-aws-logs.sh
joaquinfilipic-dynatrace: I kept it like that because in the Azure forwarder the optional usage is only indicated in that first line of the help, with those brackets, and enough with it. There were already optional arguments that have no explicit "Optional" written. But never mind, minor change, so I can add it.

equardo: As far as I can see if there is no "Optional" written then it says something about "by default" or "this defaults to". Yeah, it is minor though.

equardo (approved):

Dawid-Kaszubski (approved):

PR #88, opened 2023-08-16
Added timeout for CloudWatch call

joaquinfilipic-dynatrace:


nikita-paskal-dynatrace (approved):

src/logs/self_monitoring/sfm.py
```diff
@@ -155,7 +156,8 @@ def _generate_metrics(self):
 
     def push_sfm_to_cloudwatch(self):
         metrics = self._generate_metrics()
-        cloudwatch = boto3.client('cloudwatch')
+        config = Config(connect_timeout=5, retries={'max_attempts': 1})
```
NematulloKozimov, line 159: If exception happens when request timeouts, we still have the same issue because max_attempts was lambda function configuration, not the request itself

PR #89, opened 2023-12-19
Bugfix/pclouds 3435 aws log forwarder handle incorrect event schema gracefully

emjot13-dynatrace:
Small PR to explicitly raise an exception, when the received event cannot be parsed.

### DIFF
--- a/src/logs/input_records_decoder.py
+++ b/src/logs/input_records_decoder.py
@@ -83,3 +83,12 @@ def sfm_report_kinesis_records_age(records, context):
     except Exception as e:
         log_error_with_stacktrace(e, "Failed to calculate Kinesis Record Delay Self Monitoring",
                                   "sfm-record-delay-calc-exception")
+
+class BadSchemaError(Exception):
+    '''BadSchemaError is raised when the received logs do not meet the expected format.'''
+
+    def __init__(self, erroneous_field: str) -> None:
+        message = f"""Lambda was called with an event of unrecognized schema.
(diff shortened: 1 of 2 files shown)


src/logs/input_records_decoder.py
```diff
@@ -83,3 +83,12 @@ def sfm_report_kinesis_records_age(records, context):
     except Exception as e:
         log_error_with_stacktrace(e, "Failed to calculate Kinesis Record Delay Self Monitoring",
                                   "sfm-record-delay-calc-exception")
+
+class BadSchemaError(Exception):
+    '''BadSchemaError is raised when the received logs do not meet the expected format.'''
+
+    def __init__(self, erroneous_field: str) -> None:
+        message = f"""Lambda was called with an event of unrecognized schema.
+        Make sure you have configured everything correctly. Unexpected problem happened when trying to parse field: {erroneous_field}
```
joaquinfilipic-dynatrace, line 92: Minor thing, but "everything" seems too broad :).
Could a suggestion to check the log groups be used instead? Or to check where the logs are configured to come from, something like that.

emjot13-dynatrace: thanks, it's a good point, fixed in 66c612d2ba7d10282db20ef9f426e17d4da5cb93

emjot13-dynatrace:
After the changes, the log in CloudWatch looks like this:

image

joaquinfilipic-dynatrace (approved):

PR #94, opened 2024-01-10
Preserve original log source account id

aseppala:
This pull request fixes a problem in a cross account setup where logs are shipped from other accounts than where the DynatraceAWSLogsForwarder firehose is deployed. Instead of using the account id of the kinesis batch, pick the account from the "owner" field of the original record and use it for the log metadata.

This enables setting up a centralised log collection on dedicated AWS accounts and avoids the need to deploy this infrastructure on every account separately.

### DIFF
--- a/src/logs/transformation.py
+++ b/src/logs/transformation.py
@@ -27,6 +27,7 @@ metadata_engine = MetadataEngine()
 class RecordMetadata:
     log_group: str
     log_stream: str
+    account_id: str
 
 
 def extract_dt_logs_from_single_record(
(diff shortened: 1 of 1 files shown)


ws-antti-seppala:
Thank you. However, you didn't really merge PR. I would have expected some comments if this PR was not ready to be merged as is and that I could have addressed your needs. Now my contribution is not recorded which is a shame.

aseppala:
Thank you. However, you didn't really merge PR. I would have expected some comments if this PR was not ready to be merged as is and that I could have addressed your needs. Now my contribution is not recorded which is a shame.

NematulloKozimov:
Yes, you are right. The reason for not merging the PR directly is not because we think PR was not good enough. You provided legit solution. The reason why we could not merge your PR directly is because travis build does not run (due to our internal policy - security reasons) if someone outside Dynatrace-OSS creates PR. Therefore, preventing us from merging PRs (created by someone outside the org). 

However, we made sure that your name is mentioned in the release notes: https://github.com/dynatrace-oss/dynatrace-aws-log-forwarder/releases/tag/release-0.7.11

Thank you for your understanding

aseppala:
Thank you for the explanation. In future, you could cherry-pick the useful commits from outside collaborators and include them in your PRs.

PR #100, opened 2025-01-15
Update and rename .travis.yml to .travis.yml.disable

dtwwolinski:
Disable Travis builds


dtwwolinski:
Removed Cron Jobs every 2nd day.

